PR #14, opened 2017-07-28
implemented all single-objective functions

Carl-K:
I finished implementing all the functions.  I also looked at the handy PEP8 document you sent me and I took suggestions from it.

The only function that was tricky was levi_func(x).  If you have a better implementation suggestion I'd be happy to redo it or anything else you'd like me to improve.


ljvmiranda921:
Hi! Good job! I will just review your code then I will get back to you 👍

PR #20, opened 2017-08-29
Add grid_search

SioKCronin:
Hi LJ,

I put together a simple grid_search and test for the PSO parameter optimization. 

I'm thinking it might make sense to create a MockOptimizer() class to test the method directly, independent of the pyswarms optimizers, but wanted to get your thoughts on that before building it. 

Also, I'm not sure if it's confusing that I've required `objective_func` and `iters` to be set up front when creating the GridSearch instance, but I think this might be clearer than requiring them to be passed at the time of the optimize call. Let me know if you'd prefer the latter, and I can change that. 

Once we confirm this is on a good path, I can follow the same general approach to build the random_search, and then write up a Jupyter notebook showing them both in context. 

I look forward to hearing your thoughts!

Warmly,

Siobhán

### DIFF
--- a/pyswarms/single/local_best.py
+++ b/pyswarms/single/local_best.py
@@ -110,7 +110,7 @@ class LocalBestPSO(SwarmBase):
             a tuple of size 2 where the first entry is the minimum velocity
             and the second entry is the maximum velocity. It 
             sets the limits for velocity clamping. 
-        options : dict with keys :code:`{'c1', 'c2', 'k', 'p'}`
+        options : dict with keys :code:`{'c1', 'c2', 'w', 'k', 'p'}`
             a dictionary containing the parameters for the specific 
             optimization technique
                 * c1 : float
--- a/pyswarms/utils/search/grid_search.py
+++ b/pyswarms/utils/search/grid_search.py
@@ -1,0 +1,153 @@
+# -*- coding: utf-8 -*-
+
+"""
+Hyperparameter grid search.
+
+Compares the relative performance of hyperparameter value combinations in
+reducing a specified objective function.
+
+For each hyperparameter, user can provide either a single value or a list
+of possible values, and their cartesian product is taken to produce a grid
+of all possible combinations. These combinations are then tested to produce
+a list of objective function scores. The default of the optimize method
+returns the hyperparameters that yield the minimum score, yet maximum score
+can also be evaluated.
+
+Parameters
+----------
+* c1 : float
+    cognitive parameter
+* c2 : float
+    social parameter
+* w : float
+    inertia parameter
+* k : int
+    number of neighbors to be considered. Must be a
+    positive integer less than `n_particles`
+* p: int {1,2}
+    the Minkowski p-norm to use. 1 is the
+    sum-of-absolute values (or L1 distance) while 2 is
+    the Euclidean (or L2) distance.
+
+>>> options = {'c1': [1, 2, 3],
+               'c2': [1, 2, 3],
+               'w' : [2, 3, 5],
+               'k' : [5, 10, 15],
+               'p' : 1}
+>>> g = GridSearch(LocalBestPSO, n_particles=40, dimensions=20,
+                   options=options, objective_func=sphere_func, iters=10)
+>>> best_score, best_options = g.optimize()
+>>> best_score
+301.418815268
+>>> best_options['c1']
+1
+>>> best_options['c2']
+1
+"""
+
+import operator as op
+import itertools
+import numpy as np
+
+class GridSearch(object):
+    """Exhaustive search of optimal performance on selected objective function
+    over all combinations of specified hyperparameter values."""
+
+    def __init__(self, optimizer, n_particles, dimensions, options,
+                 objective_func, iters, bounds=None, velocity_clamp=None):
+        """Initializes the GridSearch.
+
+        Attributes
+        ----------
+        optimizer: PySwarms class
+            either LocalBestPSO or GlobalBestPSO
+        n_particles : int
+            number of particles in the swarm.
+        dimensions : int
+            number of dimensions in the space.
+        options : dict with keys :code:`{'c1', 'c2', 'w', 'k', 'p'}`
+            a dictionary containing the parameters for the specific
+            optimization technique
+                * c1 : float
+                    cognitive parameter
+                * c2 : float
+                    social parameter
+                * w : float
+                    inertia parameter
+                * k : int
+                    number of neighbors to be considered. Must be a
+                    positive integer less than :code:`n_particles`
+                * p: int {1,2}
+                    the Minkowski p-norm to use. 1 is the
+                    sum-of-absolute values (or L1 distance) while 2 is
+                    the Euclidean (or L2) distance.
+        objective_func: function
+            objective function to be evaluated
+        iters: int
+            number of iterations
+        bounds : tuple of np.ndarray, optional (default is None)
+            a tuple of size 2 where the first entry is the minimum bound
+            while the second entry is the maximum bound. Each array must
+            be of shape :code:`(dimensions,)`.
+        velocity_clamp : tuple (default is :code:`None`)
+            a tuple of size 2 where the first entry is the minimum velocity
+            and the second entry is the maximum velocity. It
+            sets the limits for velocity clamping.
+        """
+
+        # Assign attributes
+        self.optimizer = optimizer
+        self.n_particles = n_particles
+        self.dims = dimensions
+        self.options = options
+        self.bounds = bounds
+        self.vclamp = velocity_clamp
+        self.objective_func = objective_func
+        self.iters = iters
+
+    def generate_grid(self):
+        """Generates the grid of all hyperparameter value combinations."""
+
+        #Extract keys and values from options dictionary
+        params = self.options.keys()
+        items = [x if type(x) == list \
+                 else [x] for x in list(zip(*self.options.items()))[1]]
+
+        #Create list of cartesian products of hyperparameter values from options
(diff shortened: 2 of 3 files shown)


ljvmiranda921:
Hey, thanks a lot @SioKCronin! I'll check this right away once I arrive home 👍

ljvmiranda921:
Hi @SioKCronin , great work!

I realized that using the `operator` module is practical. I admit I haven't used much from that standard lib before. The implementation of the interface (from the example in the docstring) is really nice, much better than how I would implement it.

As for your questions,

1. I think it's fine to set upfront the objective function and number of iterations whenever they're doing the search. In my opinion, this forces the user to "create separate environments" for what they are searching for. I'm more for making these "search environments" as atomic as possible. My wary is that if we put them on the "optimize" method, some inadvertent state changes may occur.

2. Creating a mock-up class for an optimizer during testing is a good idea 👍, but how do you plan to implement it? Can this be done by creating a setUp fixture instead? 😄 

Just a few comments in terms of style and structure:

1. __On structure.__ I think it would be better if we put all search utilities inside a `pyswarms.utils.search` module, this means we add another directory `/search` then just put all GridSearch, RandomSearch and other mixins inside that. The `pyswarms.utils.functions` is more for objective functions (single objective, multi-obj, etc). Perhaps my use of `functions` as a module name is quite confusing? 😕 

2. __A few notes on docstrings.__ For `grid_search.py`, we can move the whole docstring (the one with the example) __from__ the class `GridSearch` __into__ the higher level docstring before you imported the modules. 

3. __More on docstrings.__ It will also be better (and will render nicely in the documentation once built) if we add some docstrings in our methods. For example, in `__init__`, we can set an `Attributes` header to describe the required attributes when initializing the class. In this library, we are following the [numpydocs](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) style guide. But you can just check the docstrings in one of the modules (like `global_best.py` or `local_best.py`) as a reference. It may look like work but it's enjoyable!

4. __Just a matter of style.__ I think from Python 2.2, the old-style declaration of classes is not recommended anymore. So you can define your `GridSearch` as `class GridSearch(object)`. You can refer to these references (Warning: it's a rabbit hole! [(1)](https://stackoverflow.com/questions/4015417/python-class-inherits-object)[(2)](https://docs.python.org/release/2.2.3/whatsnew/sect-rellinks.html))

5. __Private attributes and methods.__ For consistency with other classes, I think it's fine to just remove the underscore in our attributes in the `GridSearch` class. It's fine to use `self.optimizer` than `self._optimizer` and etc. 

Lastly, it may be helpful, when you go about writing `RandomSearch` to check if there are areas of code that are redundant. It may be more efficient to create a `BaseSearch` class where `GridSearch` and `RandomSearch` will inherit, only overriding the `optimize()` method where they both differ.

All in all, this is great work and I'm really thankful for your help!

pyswarms/single/local_best.py
```diff
@@ -110,7 +110,7 @@ def __init__(self, n_particles, dimensions, options, bounds=None, velocity_clamp
             a tuple of size 2 where the first entry is the minimum velocity
             and the second entry is the maximum velocity. It 
             sets the limits for velocity clamping. 
-        options : dict with keys :code:`{'c1', 'c2', 'k', 'p'}`
+        options : dict with keys :code:`{'c1', 'c2', 'w', 'k', 'p'}`
```
ljvmiranda921: Thank you for catching this!

pyswarms/utils/functions/grid_search.py
```diff
@@ -0,0 +1,102 @@
...
+a list of objective function scores. The default of the optimize method
+returns the hyperparameters that yield the minimum score, yet maximum score
+can also be assessed.
+"""
+
+import operator as op
+import itertools
+import numpy as np
+
+class GridSearch():
```
ljvmiranda921: I think it would be better to use the new-style introduced in Python 2.2 for declaring classes.

```python
# Old-style
class GridSearch:
    pass

# New-style
class GridSearch(object):
    pass
```

pyswarms/utils/functions/grid_search.py
```diff
@@ -0,0 +1,102 @@
...
+                   'k' : [5, 10, 15],
+                   'p' : 1}
+    >>> g = GridSearch(LocalBestPSO, n_particles=40, dimensions=20,
+                       options=options, objective_func=sphere_func, iters=10)
+    >>> best_options = g.optimize()
+    >>> best_options['c1']
+    1
+    >>> best_options['c2']
+    1
+    """
```
ljvmiranda921: You can move this at the top-level docstring before you begin importing the modules

```python
"""Hyperparameter grid search

All descriptions, examples, and what-nots goes here
"""

import operator as op
import itertools
import numpy as np
# and so on
```

pyswarms/utils/functions/grid_search.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __init__(self, optimizer, n_particles, dimensions, options,
+                 objective_func, iters, bounds=None, velocity_clamp=None):
+        """Initializes with dict of key, parameter value pairs."""
```
ljvmiranda921: We can improve upon the docstring by following [numpydocs](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt). Or you can check the docstring implementations in `pyswarms.single.LocalBestPSO`

Sample:

```python
"""
Attributes
-----------
optimizer : method
    <description goes here>
n_particles : int
    <description goes here>
"""

```

pyswarms/single/local_best.py
ljvmiranda921: Thanks for catching this! 👍

pyswarms/utils/functions/grid_search.py
```diff
@@ -0,0 +1,102 @@
...
+        self._optimizer = optimizer
+        self._n_particles = n_particles
+        self._dims = dimensions
+        self._options = options
+        self._bounds = bounds
+        self._vclamp = velocity_clamp
+        self._objective_func = objective_func
+        self._iters = iters
```
ljvmiranda921: I think it's fine to make these attributes "public" since its only the class that will access it for now. Just a matter of consistency with the other classes that we have. It's fine to remove the underscores. 😄

pyswarms/utils/functions/grid_search.py
```diff
@@ -0,0 +1,102 @@
...
+        #Return list of dicts for all hyperparameter value combinations
+        return [dict(zip(*[params, list(x)])) for x in list_of_products]
+
+    def optimize(self, maximum=False):
+        best_options = None
+        params = self._options.keys()
+        def generate_score(options):
+            f = self._optimizer(self._n_particles, self._dims, options,
+                                self._bounds, self._vclamp)
+            return f.optimize(self._objective_func, self._iters)[0]
```
ljvmiranda921: I haven't checked if there are state-effects by adding a closure inside `optimize()`. Hmmm...will it change if we just convert `generate_report()` into a private function `_generate_report()` outside of `optimize()`? 

It's just a matter of personal style, it still depends on you if you keep this or not. I think closures are good as well. 👍

pyswarms/utils/functions/grid_search.py
SioKCronin: Yes, I agree.

pyswarms/utils/functions/grid_search.py
SioKCronin: That makes total sense and looks better too!

pyswarms/utils/functions/grid_search.py
SioKCronin: Great. I brought in the descriptions from LocalBestPSO.

pyswarms/utils/functions/grid_search.py
SioKCronin: I'm a big fan of consistency :+1:, and have removed the underscores.

pyswarms/utils/functions/grid_search.py
SioKCronin: I moved `generate_score` out, and I think this reads better and is conceptually more in line with how both `generate_grid` and `generate_score` behave in relationship to `optimize` (providing functionality for the report generation). 

I like your use of the word 'report', and it has me wondering if it could be used, or something like it, instead of 'optimize'. That might help make it clear that we are reporting performance of an optimizer in this method, yet that reporting is actually 'optimizing' anything, just running the optimizer with different parameters. Maybe `generate_report`?

pyswarms/single/local_best.py
SioKCronin: No problem! 😄 

It's been great to read the documentation on this project, as it's super clear, and I'm glad you're setting a nice standard for that.

SioKCronin:
This sounds great. I'm traveling, 🛩 , but will have time tomorrow to get this spruced up. Cheers!

ljvmiranda921:
Okay, @SioKCronin , enjoy and take care!!

SioKCronin:
Hi @ljvmiranda921 ! I apologize for the delay. I've rebased and squashed my commits (thanks for the great tutorials on that). Definitely let me know if there is anything I missed.

ljvmiranda921:
Hello @SioKCronin ! All checks passed! Congratulations and thank you for your contribution! 🥂 😃 🎉 

For the other hyperparameter search tools like RandomSearch (or the BaseSearch class), just open another Pull Request!

SioKCronin:
Awesome! Yes, I'm excited to get RandomSearch up and running, and will write that this week. 

Thanks for your clear and helpful reviews! 🙌

PR #28, opened 2017-09-20
Implement assertions() in search #27

jazcap53:
… (#27)

Check name of optimizer member in SearchBase.
Check type of n_selection_iters member in RandomSearch.
Add self.assertions() line to __init__() in each of Searchbase, GridSearch, and RandomSearch.
Add call to super().assertions() in RandomSearch.

Author: jazcap53
Email: andrew.jarcho@gmail.com

### DIFF
--- a/pyswarms/utils/search/base_search.py
+++ b/pyswarms/utils/search/base_search.py
@@ -13,6 +13,20 @@ import operator as op
 from past.builtins import xrange
 
 class SearchBase(object):
+
+    def assertions(self):
+        """Assertion method to check :code:`optimizer` input.
+
+        Raises
+        ------
+        TypeError
+            When :code:`optimizer` does not have an `'optimize'` attribute.
+        """
+        # Check type of optimizer object
+        if not hasattr(self.optimizer, 'optimize'):
+            raise TypeError('Parameter `optimizer` must have an '
+                            '`\'optimize\'` attribute.')
+
     def __init__(self, optimizer, n_particles, dimensions, options,
                  objective_func, iters,
                  bounds=None, velocity_clamp=None):
@@ -65,6 +79,8 @@ class SearchBase(object):
         self.vclamp = velocity_clamp
         self.objective_func = objective_func
         self.iters = iters
+        # Invoke assertions
+        self.assertions()
 
     def generate_score(self, options):
         """Generates score for optimizer's performance on objective function.
@@ -73,7 +89,7 @@ class SearchBase(object):
         ----------
 
         options: dict
-            a dict of 5 hyperparameter values ('c1', 'c2', 'w', 'k', 'p' 
+            a dict with the following keys: {'c1', 'c2', 'w', 'k', 'p'}
         """
 
         #Intialize optimizer
@@ -91,9 +107,9 @@ class SearchBase(object):
         ----------
 
         maximum: bool
-            a bool defaulting to False, returning the minimum value for the 
+            a bool defaulting to False, returning the minimum value for the
             objective function. If set to True, will return the maximum value
-            for the objective function. 
+            for the objective function.
         """
 
         #Assign parameter keys
--- a/pyswarms/utils/search/grid_search.py
+++ b/pyswarms/utils/search/grid_search.py
@@ -54,6 +54,8 @@ class GridSearch(SearchBase):
         super(GridSearch, self).__init__(optimizer, n_particles, dimensions, options,
                 objective_func, iters, bounds=bounds,
                 velocity_clamp=velocity_clamp)
+        # invoke assertions
+        self.assertions()
 
     def generate_grid(self):
         """Generates the grid of all hyperparameter value combinations."""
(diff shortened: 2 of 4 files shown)


pyswarms/utils/search/base_search.py
```diff
@@ -13,6 +13,21 @@
...
+        ------
+        TypeError
+            When :code:`optimizer` is not named LocalBestPSO or
+                GlobalBestPSO.
+        """
+        # Check type of optimizer object
+        if self.optimizer.__name__ not in ('LocalBestPSO', 'GlobalBestPSO'):
+            raise TypeError('Parameter `optimizer` must be of type '
+                            'LocalBestPSO or GlobalBestPSO')
+
```
ljvmiranda921: This is good for now and I can accept this in its current form. 😄 👍 

But I'm wondering how can we scale this when more search methods will be added. For example, Binary PSO is a single-objective method, and I think the same search methods are applicable to it. 

The same will go perhaps in multi-objective and constrained optimization. Since the search methods are "independent" whatever optimization method we are using. Meaning, it doesn't really care what kind of problem we are trying to solve, as long as we have an `optimize` and `optimizer` supplied as params. Then it's possible to apply the same search methods for these kind of techniques in the future. What do you think @SioKCronin?

I'm thinking of instead checking if the object passed in the `optimizer`  argument belongs to the `pyswarms.base` module. __OR__ check if the object passed in the `optimizer` method has an `optimize` attribute (thus using a `hasattr` as a checker). What do you think @jazcap53?

pyswarms/utils/search/base_search.py
```diff
@@ -73,7 +90,7 @@ def generate_score(self, options):
         ----------
 
         options: dict
-            a dict of 5 hyperparameter values ('c1', 'c2', 'w', 'k', 'p' 
+            a dict of 5 hyperparameter values ('c1', 'c2', 'w', 'k', 'p'
```
ljvmiranda921: Is a parenthesis missing here? 😄 ❓ 
I guess we can also rephrase this as:
```python
"""
a dict with the following keys: {'c1 ', 'c2', 'w', 'k', 'p'}
"""
```

pyswarms/utils/search/base_search.py
jazcap53: Hi @ljvmiranda921,
I'll of course make the changes you suggest with pleasure. :smile:
> I'd appreciate it if you can rebase your commit similar to the title of your pull request  

I had trouble getting git to accept a commit title with spaces in it. I'll try using `\<space>` instead of `<space>`, and if that doesn't work I'll try using `%20`.

> I'm wondering how can we scale this when more search methods will be added

I was wondering the same thing. I had trouble implementing it, though, since the `optimizer` parameter is a class rather than an instance -- I'll research it further and ask help if I can't find a solution.

> Is a parenthesis missing here?

Sure enough. I was afraid to change that line, though -- it's not mine. I just appears in the diff because my vim is set to eliminate trailing spaces on save of a .py file. I'll rephrase it according to your suggestion.

I'm enjoying this very much...  :+1:

pyswarms/utils/search/base_search.py
jazcap53: Hi @ljvmiranda921 

> Is a parenthesis missing here?

Sure enough. :smile: I was afraid to change that line, though -- it's not mine. I just appears in the diff because my vim is set to eliminate trailing spaces on save of a .py file. I'll rephrase it according to your suggestion.

pyswarms/utils/search/base_search.py
jazcap53: @ljvmiranda921   Figured out that I have to use `issublcass()`. Simple (embarrassing). :smile:

jazcap53: @ljvmiranda921    o.i.c.   `if not hasattr(self.optimizer, 'optimize'):`
I'll use that unless you would prefer something else  :smile:

jazcap53:
Hi @ljvmiranda921 !

I've followed all instructions, I hope correctly. Google and Stack Overflow have been getting a lot of my business lately! :smile:

If there's anything else I need to do, please of course let me know. 

Next I plan to study up on Travis CI and write some tests for the 'Go the extra mile?' portion of the Issue.

:+1:

ljvmiranda921:
Hi @jazcap53 ! All is good! I will merge this now 👍 

Congratulations for your first contribution!!! :D 

For the extra mile, no need to actually study travis-ci because I have already set that one up beforehand. You just need to write the unit tests, and travis-ci will automatically run your tests using `discover`. 😄 

So in case that one of the tests didn't pass, it will send a notification here in Github and also in my e-mail. In the 'Extra mile', you will just write the unit tests that travis-ci will run. 😄

jazcap53:
@ljvmiranda921 
Yesssssss! :smile: :+1:
Thanks so much for all your help. Looking forward to doing the tests!

PR #31, opened 2017-09-23
Install ffmpeg to run animation tests

ljvmiranda921:
This pull request aims to install the ffmpeg in travis-ci. This
is to facilitate the success of checks that use the `matplotlib.animation.FuncAnimation`
class. The modifications are done in the `.travis.yml` file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,9 @@ deploy:
     repo: ljvmiranda921/pyswarms
     tags: true
 install: 
-  - "pip install -U tox-travis"
+  - sudo apt-get update
+  - sudo apt-get install libav-tools
+  - pip install -U tox-travis
 language: python
 python:
   - 3.6
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ requirements = [
     'scipy>=0.17.0',
     'numpy>=1.13.0',
     'matplotlib>=1.3.1',
+    'mock==2.0.0',
 ]
 
 setup_requirements = [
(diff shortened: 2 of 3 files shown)


ljvmiranda921:
That's so sad, what if I just install `ffmpeg` without creating a `test-environment`? Either way, I will still be checking it via `tox`. Also, it is weird that the HTTP errors are now gone.

ljvmiranda921:
It turns out that there is an issue between the tox virtualenv and the environments created by
conda. I will resolve this by installing virtualenv via conda.

#### Relevant Discussions 
- [BitBucket Actual Issue](https://bitbucket.org/hpk42/tox/issues/273/support-conda-envs-when-using-miniconda)
- [sdayal's answer in StackOverflow](https://stackoverflow.com/questions/30555943/is-it-possible-to-use-tox-with-conda-based-python-installations)

ljvmiranda921:
It seems that `ffmpeg` has already moved into another package whatsoever. I am now installing another toolchain called `libav-tools`

#### Relevant Discussions
- [Software installation of ffmpeg](https://askubuntu.com/questions/432542/is-ffmpeg-missing-from-the-official-repositories-in-14-04)

ljvmiranda921:
So this is working already, just need to install `mock`. My bad.

ljvmiranda921:
Now there is another problem with respect to the display errors in `_tkinter.TclError`. I tried
solving this by changing the environment variables for the display.

#### Relevant discussions
- [ajgriese's StackOverflow answer](https://stackoverflow.com/questions/37604289/tkinter-tclerror-no-display-name-and-no-display-environment-variable)
- [matplotlib's travis.yml](https://github.com/matplotlib/matplotlib/blob/master/.travis.yml)

PR #34, opened 2017-09-25
Update binary.py

CPapadim:
Hi, I'm the reporter of issue 33

I think the issue is that rather than moving toward and returning the historical social / global best, you are moving toward and returning the current social / global best.  I think for PSO you always need to be moving to historical and not current positions whether its the personal or the social / global best.

I've made some updates to fix the issue.

I tested on my code, and with my updates every iteration I either get the same 'best' position I had in the previous iteration, or a better one.

I'm not super familiar with this package's codebase so this request likely needs review and testing.

### DIFF
--- a/pyswarms/discrete/binary.py
+++ b/pyswarms/discrete/binary.py
@@ -171,9 +171,9 @@ class BinaryPSO(DiscreteSwarmBase):
             # Obtain the indices of the best position for each
             # neighbour-space, and get the local best cost and
             # local best positions from it.
-            nmin_idx = self._get_neighbors(current_cost)
-            self.best_cost = current_cost[nmin_idx]
-            self.best_pos  = self.pos[nmin_idx]
+            nmin_idx = self._get_neighbors(pbest_cost)
+            self.best_cost = pbest_cost[nmin_idx]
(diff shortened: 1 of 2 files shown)


ljvmiranda921:
Hey, thanks for the fix! I can accept this already, but would you like to compare your fix to [mine](https://github.com/ljvmiranda921/pyswarms/pull/35)?

CPapadim:
As we discussed, I think my fix keeps neighbors working properly.  So that we're only using nearby particles for the social best rather than all of them.  I can update local_best.py similarly

ljvmiranda921:
Okay then, I'll just wait for the commit to be pushed :)

ljvmiranda921:
Nice, this is good to go 👍 

Just one last logistical request. Normally I ask our contributors to rebase commits into 1 or 2. For this, probably one commit for the `binary.py` and another for `local_best.py`. And also, we follow this commit format:

```
Short and sweet imperative title (#32)

Short description that describes why the commit was done

Author: <your-github-username>
Email: <your-contact-email>
```

That's all, after this we can merge already. Thanks a lot!

PR #65, opened 2017-11-21
Fix offset shape for 2D animation

emascot:
Offset shape should be (N,2). The bug can be seen when using the Beale function (which had a typo for the global minimum).

### DIFF
--- a/pyswarms/utils/environments/plot_environment.py
+++ b/pyswarms/utils/environments/plot_environment.py
@@ -381,8 +381,8 @@ class PlotEnvironment(object):
             iterable of artists
         """
         current_pos = data[i]
-        x, y = current_pos[:, idx.x], current_pos[:, idx.y]
-        plot.set_offsets([x, y])
+        xy = current_pos[:, (idx.x, idx.x)]
+        plot.set_offsets(xy)
         return plot,
 
     def _animate3D(self, i, data, plot, idx):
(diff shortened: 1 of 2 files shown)


pyswarms/utils/environments/plot_environment.py
```diff
@@ -381,8 +381,8 @@ def _animate2D(self, i, data, plot, idx):
             iterable of artists
         """
         current_pos = data[i]
-        x, y = current_pos[:, idx.x], current_pos[:, idx.y]
-        plot.set_offsets([x, y])
+        xy = current_pos[:, (idx.x, idx.y)]
+        plot.set_offsets(xy)
         return plot,
```
ljvmiranda921: I see, thanks for catching this!

pyswarms/utils/functions/single_obj.py
```diff
@@ -141,7 +141,7 @@ def beale_func(x):
     """Beale objective function.
 
     Only takes two dimensions and has a global minimum at
-    :code:`f([0,3.5])` Its domain is bounded between :code:`[-4.5, 4.5]`
+    :code:`f([3,0.5])` Its domain is bounded between :code:`[-4.5, 4.5]`
```
ljvmiranda921: Oops, thanks again for this fix!

emascot:
Hi @ljvmiranda921. I squashed and reworded the commit. And yes, those are both me.

ljvmiranda921:
Merged! Thanks a lot @emascot !

PR #93, opened 2018-04-05
Add guess and acceptable tolerance of objective function evaluation features

mamadyonline:
Added two new features which are:

1. Provide the possibility for the user to set manually a starting point for both global and local PSO.
2. Provide the possibility for the user to specify manually a tolerance in the objective function evaluation.

These features are available for many optimization algorithms such as those used in the scipy optimize library. I think they are useful for instance when we want to compare PSO with Nelder-Mead (the version under scipy optimize), with those two attributes, we can set them on equal footing.

Note: Had to close my last PR because it does not pass the travis test, due to the test of tolerance feature. The behavior is normal since I am doing a simple test (in fact I am betting on Local PSO to finish before the maximum iteration which is not guaranteed) on the fact that the optimizer would finish before the maximum number of iteration, but somehow it failed.

### DIFF
--- a/pyswarms/base/base_single.py
+++ b/pyswarms/base/base_single.py
@@ -82,12 +82,12 @@ class SwarmBase(object):
                 raise ValueError('Make sure that velocity_clamp is in the '
                                  'form (min, max)')
 
-        # Check setting of guess
-        if self.guess is not None:
-            if not isinstance(self.guess, list):
-                raise TypeError('Parameter `guess` must be a list')
-            if not len(self.guess) == self.dimensions:
-                raise IndexError('Parameter `guess` must be the same shape '
+        # Check setting of init_pos
+        if self.init_pos is not None:
+            if not isinstance(self.init_pos, list):
+                raise TypeError('Parameter `init_pos` must be a list')
+            if not len(self.init_pos) == self.dimensions:
+                raise IndexError('Parameter `init_pos` must be the same shape '
                                  'as dimensions.')
 
         # Required keys in options argument
@@ -119,7 +119,7 @@ class SwarmBase(object):
             logging.basicConfig(level=default_level)
 
     def __init__(self, n_particles, dimensions, options,
-                 bounds=None, velocity_clamp=None, guess=None, ftol=-np.inf):
+                 bounds=None, velocity_clamp=None, init_pos=None, ftol=-np.inf):
         """Initializes the swarm.
 
         Creates a :code:`numpy.ndarray` of positions depending on the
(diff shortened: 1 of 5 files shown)


ljvmiranda921:
Awesome, I'll review this in a while. Thank you very much!

ljvmiranda921:
Sorry for the delay, I'll do this on Tuesday after my thesis defense. Wish me luck!

pyswarms/base/base_single.py
```diff
@@ -82,6 +82,14 @@ def assertions(self):
                 raise ValueError('Make sure that velocity_clamp is in the '
                                  'form (min, max)')
 
+        # Check setting of guess
```
ljvmiranda921: Just wondering about the use of the word `guess`. We are setting for the initial position of the particles, right? Instead of having it be defined by our bounds. I'd suggest we go for `init_pos` or something that is more explicit. We can update the documentation to mention the word "guess" instead.

tests/optimizers/test_global_best.py
```diff
@@ -165,5 +177,14 @@ def test_velocity_history_size(self):
         velocity_hist = self.optimizer.get_velocity_history
         self.assertEqual(velocity_hist.shape, (1000, 10, 2))
 
+    def test_ftol_effect(self):
+        """Check if setting ftol breaks the optimization process
```
ljvmiranda921: This is a good test. Thanks!

tests/optimizers/test_local_best.py
```diff
@@ -123,6 +123,18 @@ def test_vclamp_minmax_fail(self):
         with self.assertRaises(ValueError):
             optimizer = LocalBestPSO(5,2,velocity_clamp=velocity_clamp, options=self.options)
 
+    def test_guess_type_fail(self):
+        """Tests if exception is thrown when guess is not a list."""
+        guess = (0.1,1.5)
+        with self.assertRaises(TypeError):
+            optimizer = LocalBestPSO(5,2,guess=guess,options=self.options)
```
ljvmiranda921: Please put a line break between 130 and 131

tests/optimizers/test_global_best.py
```diff
@@ -100,6 +100,18 @@ def test_vclamp_minmax_fail(self):
         with self.assertRaises(ValueError):
             optimizer = GlobalBestPSO(5,2,velocity_clamp=velocity_clamp,options=self.options)
 
+    def test_guess_type_fail(self):
+        """Tests if exception is thrown when guess is not a list."""
+        guess = (0.1,1.5)
+        with self.assertRaises(TypeError):
+            optimizer = GlobalBestPSO(5,2,guess=guess,options=self.options)
+    def test_guess_shape_fail(self):
```
ljvmiranda921: Please put a line break between 107 and 108 :+1:

mamadyonline:
I'll check them today when I am back home.  Good point indeed about the name of "guess" for the initial position. I will make the changes and submit a new PR.
Thanks for the comments.

mamadyonline:
I tried hard fixing the commits :(
Let me know of any remarks about the new changes.

ljvmiranda921:
Awesome @mamadyonline , I have squashed the commits instead and is now part of the master branch! Thank you so much! :+1:

mamadyonline:
It was a pleasure! You are easy to work with ;)

Issue #110, opened 2018-05-09
Returning an empty subset as the best subset for feature selection


ahcantao:
* PySwarms version: 0.1.9
* Python version: 3.5.4
* Operating System: windows 10

### Description

The code is returning an empty subset of attributes as the best subset selection. It is even returning the cost of the subset with only zeros. Even with a minor change on the exemple code it is possible to replicate this output.

There is a checking condition on the code to avoid empty subsets but it still outputs empty as the best subset.
if np.count_nonzero(m) == 0: 
        #if the particle subset is only zeros, get the original set of attributes
        X_subset = X

Describe what you were trying to get done.
Return a non-empty subset of attributes (a binary list containing at least one element 1)

### What I Did
I ran the "**feature_subset_selection**" notebook (https://github.com/ljvmiranda921/pyswarms/blob/master/examples/feature_subset_selection.ipynb)  up to the **# Perform optimization** point. It all runs ok, but when I do a minor change, just on the number of iterations from 1000 to 10, it most of the times return an empty subset as the best subset.

The only line changed was the following:
```
cost, pos = optimizer.optimize(f, print_step=1, iters=10, verbose=2)
print(cost,pos)
```
and the output was:
```
INFO:pyswarms.discrete.binary:Iteration 1/10, cost: 0.2696
INFO:pyswarms.discrete.binary:Iteration 2/10, cost: 0.2696
INFO:pyswarms.discrete.binary:Iteration 3/10, cost: 0.2696
INFO:pyswarms.discrete.binary:Iteration 4/10, cost: 0.2696
INFO:pyswarms.discrete.binary:Iteration 5/10, cost: 0.2696
INFO:pyswarms.discrete.binary:Iteration 6/10, cost: 0.2696
INFO:pyswarms.discrete.binary:Iteration 7/10, cost: 0.2376
INFO:pyswarms.discrete.binary:Iteration 8/10, cost: 0.2376
INFO:pyswarms.discrete.binary:Iteration 9/10, cost: 0.2376
INFO:pyswarms.discrete.binary:Iteration 10/10, cost: 0.2376
INFO:pyswarms.discrete.binary:================================
Optimization finished!
Final cost: 0.2376
Best value: [ 0.000000 0.000000 0.000000 ...]

0.2376 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
```

ljvmiranda921:
Hi @ahcantao , sorry for the late reply.

[...]

Issue #111, opened 2018-05-13
Write a backend module


ljvmiranda921:
I think it's better to decompose some methods in the swarm loop into its own backend module.
Makes it easier for us to test what's happening inside.

Deps: #101 
Port all tests to pytest first before doing this.

- [x] Accomplish Issue #101 
- [x] Write a backend module
- [x] Add tests to backend module
- Update documentation about the module (make new Issue on this)

PR #115, opened 2018-05-19
Add backend module

ljvmiranda921:
Reference: #111


ljvmiranda921:
I was wrong, we should bring back the option where `velocity_clamp = None`. I thought I can go around it by simply setting `clamp = (0,1)`. The previous commit fixes that.

ljvmiranda921:
I'll probably finish everything first before I request for review :+1:  But you can comment whenever you feel

ljvmiranda921:
## TODO 

- [ ] Update BinaryPSO and DiscretePSO with new backends (might be unruly so hopefully we can also fix #110 in the process)
- [ ] Update Documentation (i.e., ReadTheDocs, some examples on how to use the backend, short example in creating your own evolutionary-loop [white-box approach)

ljvmiranda921:
Should we actually use topologies instead when computing for the best cost and position?

I'm thinking of generalizing the `compute_gbest()` and `compute_gbest_neighborhood()` methods in just a single Topology. The idea is that we have a `compute_best_particle` that changes its behavior depending on the topology present. This changes how `best_pos` and `best_cost` is computed and in turn affects the particle's velocity.

```python
best_cost, best_pos = compute_best_particle(swarm, topology='star') # For GlobalBestPSO
```

PR #116, opened 2018-05-20
Add a separate Swarm class 

ljvmiranda921:
**Reference**: #111, #115 
**Branch**: Development

The idea is to create a Swarm class that will be fed into the Optimizer class (Current our BaseClass).
This will introduce some breaking changes in the API that's why we'll only merge this into the `development` branch.

For common applications, importing `GlobalBestPSO` and calling the `optimize()` function should work as with the `master` branch. It's just that the backend is different (we're not keeping everything inside the PSO class anymore).

**Motivation**: I want to have a more solid white-box approach in building custom swarm operations. Sure, everyone can use `GlobalBestPSO` and `LocalBestPSO` in most use-cases, but it might be nice to have a set of backend modules that supports prototyping of PSO implementations.

Another motivation is that the current backend I implementend (see #115) has a lot of parameters and is a bit clunky. Keeping everything in a swarm class--defined by its position, velocity, and behavior (`options`)--might be more intuitive.

**Sample**

If we wish to update the personal best or velocity we have to do this:

```python
import pyswarms.backend as P

# Inside the for-loop...
new_pbest_pos, new_pbest_cost = P.update_pbest(pbest_pos=current_pbest_pos,
                                                                                   pbest_cost=current_pbest_cost,
                                                                                   pos=current_pos,
                                                                                   cost=current_cost)

new_velocity = P.update_velocity(velocity=current_velocity, clamp=clamp, pos=current_pos,
                                                     pbest_pos=current_pbest_pos, best_pos=best_pos, c1=c1,
                                                      c2=c2, w=w)
```

It's a bit clunky and may be un-Pythonic.
I'm thinking of having a separate swarm class (different from the Optimizer class) so that it becomes the following:

```python
import pyswarms.backend as P
from pyswarms.swarm import Swarm

# Initializing swarm from backend
swarm_class = Swarm() # Has position, velocity, behavior, pbest, fitness etc. We feed this to optimizer

# Inside the for-loop...
new_pbest_pos, new_pbest_cost = P.update_pbest(swarm_class)
new_velocity = P.update_velocity(swarm_class, clamp=clamp)
```

This allows us to have `Optimizer` classes who are good at one thing and one thing only: get a Swarm and update its parameters. The internal logic of these classes can be defined by the user in any way they want. 

At the same time, the `Swarm` class will be good at one thing and one thing only: hold swarm attributes. This includes Swarm positions, velocities, behavior, etc.

I will try update `GlobalBestPSO` with this backend and let's see what happens. Again, for common use-cases, initializing `GlobalBestPSO` and calling the `optimize()` method should work well as in the `master` branch. It's just that the backend will be different.


ljvmiranda921:
The next commits will be really ugly...

ljvmiranda921:
## TODO (Next Weekend: 5-20)
- [ ] Fix `LocalBestPSO` (update to SwarmOptimizer)
- [ ] Implement neighbors in `operators` module
- [ ] Use Swarm class for `LocalBestPSO`

PR #117, opened 2018-05-24
Add Topologies

ljvmiranda921:
The next few commits should fail, I don't have tests for it yet.


ljvmiranda921:
## TODO

- Write out the Ring topology for LocalBestPSO. The reason why your tests are failing is because it's trying to import LocalBestPSO, but that class also imports the `operator` methods that does not exist. Write out the Ring topology first.

ljvmiranda921:
Wow I solved it. :open_mouth:

Issue #130, opened 2018-06-12
Add plotting module


ljvmiranda921:
I think it is better to decouple visualization from optimization.

Before, we have a `PlotEnviroment` class that takes an optimizer instance, runs
the optimization, and plots whatever the user requires (cost history, animations, etc.).

Some problems I see: 
- What if I want to experiment with my swarm first before visualization?
- What if optimization takes a long time? Does it mean that the `PlotEnvironment` class
will repeat this whole process again?
- What if I *really* want to plot this particular history because of some interesting property
I found in this rollout? `PlotEnvironment` will just ignore that because it has its own rollout.

It's better if we have the user do the optimization first, get the histories (via the properties
in the `SwarmOptimizer` class, e.g. `my_optimizer.cost_history`) and pass that to the plotting
function (doesn't need to be contained in a class)

Rough draft:

```python
import pyswarms as ps
from pyswarms.utils.plotters import (plot_cost_history, plot_2d_trajectory)

# Setup PSO
optimizer = ps.single.GlobalBestPSO(n_particles=100, dimensions=2, options=my_options)
optimizer.optimize(f, iters=1000) # Run for 1k iterations

# Obtain history
swarm_cost_history = optimizer.cost_history  # Of shape (1000, )
swarm_pos_history = optimizer.pos_history # Of shape (1000, 100, 2)
swarm_velocity_history = optimizer.velocity_history # Of shape (1000, 100, 2)

# Plot!
# Plotters can still accept an Axis class from matplotlib

plot_cost_history(swarm_cost_history, title='My cost history')

# Plotting velocity is optional. Only plot if an arg is passed
# Still returns an animate class from matplotlib
# Contour can be turned on or off. Make a separate "backend" for this so that proper errors can
#   be passed.
plot_2d_trajectory(swarm_pos_history, swarm_velocity_history, contour=True)
```

PR #131, opened 2018-06-12
Add plotting module

ljvmiranda921:
Reference: #130


ljvmiranda921:
Perhaps we can remove the dependency on the getters? it violates Python's rule of having only one way of doing things. We can actually access the cost histories (and other things) from optimizer instance themselves.

Issue #143, opened 2018-06-25
Functionality to pass arguments to objective function


bradahoward:
**Parameterizing Objective Functions**
The ability to send parameters into the objective function will provide much more applicability for the project. 

**Solution**
Use `*args` and/or `**kwargs` to enable the passing arguments into the objective function.

ljvmiranda921:
Hey, awesome @bradahoward , let me check this on Thursday :+1:

PR #144, opened 2018-06-25
Feature parameterize obj

bradahoward:
#143 

**Implementation**
Used both `*args` and `**kwargs` so user can choose which one they prefer. Added to `GlobalBestPSO`, `LocalBestPSO` and `BinaryPSO`.  Also added an example IPython notebook to demonstrate usage.

### DIFF
--- a/pyswarms/base/base_discrete.py
+++ b/pyswarms/base/base_discrete.py
@@ -193,7 +193,7 @@ class DiscreteSwarmOptimizer(object):
         self.pos_history.append(hist.position)
         self.velocity_history.append(hist.velocity)
 
-    def optimize(self, objective_func, iters, print_step=1, verbose=1):
+    def optimize(self, objective_func, iters, args=(), kwargs={}, print_step=1, verbose=1):
         """Optimizes the swarm for a number of iterations.
 
         Performs the optimization to evaluate the objective
--- a/pyswarms/base/base_discrete.py
+++ b/pyswarms/base/base_discrete.py
@@ -193,7 +193,7 @@ class DiscreteSwarmOptimizer(object):
         self.pos_history.append(hist.position)
         self.velocity_history.append(hist.velocity)
 
-    def optimize(self, objective_func, iters, args=(), kwargs={}, print_step=1, verbose=1):
+    def optimize(self, objective_func, iters, print_step=1, verbose=1, *args, **kwargs):
         """Optimizes the swarm for a number of iterations.
 
         Performs the optimization to evaluate the objective
--- a/pyswarms/base/base_discrete.py
+++ b/pyswarms/base/base_discrete.py
@@ -193,7 +193,7 @@ class DiscreteSwarmOptimizer(object):
         self.pos_history.append(hist.position)
         self.velocity_history.append(hist.velocity)
 
-    def optimize(self, objective_func, iters, *args, print_step=1, verbose=1, **kwargs):
+    def optimize(self, objective_func, iters, print_step=1, verbose=1, **kwargs):
         """Optimizes the swarm for a number of iterations.
 
         Performs the optimization to evaluate the objective
@@ -210,6 +210,8 @@ class DiscreteSwarmOptimizer(object):
             amount of steps for printing into console.
         verbose : int (the default is 1)
             verbosity setting.
+        kwargs : dict
+            arguments for objective function
 
         Raises
         ------
--- a/pyswarms/base/base_discrete.py
+++ b/pyswarms/base/base_discrete.py
@@ -193,7 +193,7 @@ class DiscreteSwarmOptimizer(object):
         self.pos_history.append(hist.position)
         self.velocity_history.append(hist.velocity)
 
-    def optimize(self, objective_func, iters, print_step=1, verbose=1, *args, **kwargs):
+    def optimize(self, objective_func, iters, *args, print_step=1, verbose=1, **kwargs):
         """Optimizes the swarm for a number of iterations.
 
         Performs the optimization to evaluate the objective
--- a/pyswarms/base/base_single.py
+++ b/pyswarms/base/base_single.py
@@ -224,7 +224,7 @@ class SwarmOptimizer(object):
         self.pos_history.append(hist.position)
         self.velocity_history.append(hist.velocity)
 
-    def optimize(self, objective_func, iters, *args, print_step=1, verbose=1, **kwargs):
+    def optimize(self, objective_func, iters, print_step=1, verbose=1, **kwargs):
         """Optimizes the swarm for a number of iterations.
 
         Performs the optimization to evaluate the objective
@@ -241,6 +241,8 @@ class SwarmOptimizer(object):
             amount of steps for printing into console.
         verbose : int (the default is 1)
             verbosity setting.
+        kwargs : dict
+            arguments for objective function
 
         Raises
         ------
--- a/pyswarms/base/base_single.py
+++ b/pyswarms/base/base_single.py
@@ -224,7 +224,7 @@ class SwarmOptimizer(object):
         self.pos_history.append(hist.position)
         self.velocity_history.append(hist.velocity)
 
-    def optimize(self, objective_func, iters, print_step=1, verbose=1):
+    def optimize(self, objective_func, iters, args=(), kwargs={}, print_step=1, verbose=1):
         """Optimizes the swarm for a number of iterations.
 
         Performs the optimization to evaluate the objective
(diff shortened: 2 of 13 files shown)


ljvmiranda921:
Hi @bradahoward !

Awesome PR, I appreciate the unit tests and the updates in documentation. Some comments:

1. I think the best way to include `*args` and `**kwargs` is to include them directly in the `optimize()` function rather than having a positional argument with name `args=()` or `kwargs={}`:

```python
def optimize(f, iters, print_step, verbose, *args, **kwargs):
    # Stuff...
    self.swarm.current_cost = objective_func(self.swarm.position, *args, **kwargs)
    self.swarm.pbest_cost = objective_func(self.swarm.pbest_pos, *args, **kwargs)
```
Would you mind testing this one out? Using `*args` and `**kwargs` is more Pythonic and minimizes code smells :+1: . Can you do the same thing with the `search` module? If not, just revert that module to the original. I'm planning to update its API in the future.

So as far as I know, we just need to update the `base` module and the `single` module by adding the `*args` and `**kwargs` in the optimize method.

2. The `environment` module will be deprecated. No need to update or touch that. You can just revert the commit on that file so we won't have any merge conflicts down the road. :+1:

pyswarms/base/base_discrete.py
```diff
@@ -193,7 +193,7 @@ def _populate_history(self, hist):
         self.pos_history.append(hist.position)
         self.velocity_history.append(hist.velocity)
 
-    def optimize(self, objective_func, iters, print_step=1, verbose=1):
+    def optimize(self, objective_func, iters, args=(), kwargs={}, print_step=1, verbose=1):
         """Optimizes the swarm for a number of iterations.
```
ljvmiranda921: Maybe `*args` and `**kwargs` will work here instead of `args=()` and `kwargs={}`?

pyswarms/base/base_single.py
```diff
@@ -224,7 +224,7 @@ def _populate_history(self, hist):
         self.pos_history.append(hist.position)
         self.velocity_history.append(hist.velocity)
 
-    def optimize(self, objective_func, iters, print_step=1, verbose=1):
+    def optimize(self, objective_func, iters, args=(), kwargs={}, print_step=1, verbose=1):
```
ljvmiranda921: Maybe `*args` and `**kwargs` will work here instead of `args=()` and `kwargs={}`?

pyswarms/discrete/binary.py
```diff
@@ -149,7 +149,7 @@ def __init__(
         # Initialize the topology
         self.top = Ring()
 
-    def optimize(self, objective_func, iters, print_step=1, verbose=1):
+    def optimize(self, objective_func, iters, args=(), kwargs={}, print_step=1, verbose=1):
         """Optimizes the swarm for a number of iterations.
```
ljvmiranda921: Maybe `*args` and `**kwargs` will work here instead of `args=()` and `kwargs={}`?

pyswarms/single/global_best.py
```diff
@@ -153,10 +157,14 @@ def optimize(self, objective_func, iters, print_step=1, verbose=1):
         tuple
             the global best cost and the global best position.
         """
+
+        cli_print("Arguments Passed to Objective Function: \nargs: %s \nkwargs: %s\n" % (args, kwargs),
+                  verbose, 2, logger=self.logger)
+
```
ljvmiranda921: This is good, keep this. I'd suggest using:

```
cli_print("Arguments passed to objective function: \nargs: %s \nkwargs: %s\n" % (*args, **kwargs),
                  verbose, 2, logger=self.logger)
```

pyswarms/single/global_best.py
```diff
@@ -153,10 +157,14 @@ def optimize(self, objective_func, iters, print_step=1, verbose=1):
...
         for i in range(iters):
             # Compute cost for current position and personal best
-            self.swarm.current_cost = objective_func(self.swarm.position)
-            self.swarm.pbest_cost = objective_func(self.swarm.pbest_pos)
+            self.swarm.current_cost = objective_func(self.swarm.position, *args, **kwargs)
+            self.swarm.pbest_cost = objective_func(self.swarm.pbest_pos, *args, **kwargs)
             self.swarm.pbest_pos, self.swarm.pbest_cost = compute_pbest(
```
ljvmiranda921: In my opinion, this is just the modification we need. That is, to provide flexibility in the objective_function by providing an optional `*args` or `**kwargs` argument. There's no need to put `args = ()` or `kwargs = {}` in the main `optimize()` method.

pyswarms/utils/environments/plot_environment.py
```diff
@@ -104,7 +104,7 @@ def assertions(self):
                 "Missing methods in optimizer, check " "pyswarms.base module"
             )
 
-    def __init__(self, optimizer, objective_func, iters):
+    def __init__(self, optimizer, objective_func, iters, obj_args=(), obj_kwargs={}):
         """Runs the optimizer against an objective function for a number
```
ljvmiranda921: No need to touch this :+1: This will be Deprecated soon. Simply revert this to the original to avoid merge conflicts down the road.

pyswarms/utils/search/base_search.py
```diff
@@ -33,6 +33,8 @@ def __init__(
         options,
         objective_func,
         iters,
+        obj_args=(),
+        obj_kwargs={},
```
ljvmiranda921: Is it possible to avoid `obj_args=()` and `obj_kwargs={}`? If not, just revert this to the original. I'm planning to update its API soon

tests/optimizers/test_objective_func_with_kwargs.py
```diff
@@ -0,0 +1,160 @@
...
+
+# Import modules
+import pytest
+import numpy as np
+
+# Import from package
+from pyswarms.single import GlobalBestPSO, LocalBestPSO
+from pyswarms.utils.functions.single_obj import rosenbrock_func
+
+
```
ljvmiranda921: Thanks for the awesome tests! Just a few more tests to write:

1. Can you please test if an error is raised when `*args`/`**kwargs` is passed but it's not defined in the objective function itself? (e.g. I passed `c` and `d` but rosenbrock accepts `a` and `b`).
2. Can you also test if error is raised when not all `*args`/`**kwargs` are supplied? (e.g. I only passed `a` but rosenbrock accepts both `a` and `b`)
3. Test what happens if I passed more than one `*args`/`**kwargs` than necessary.

pyswarms/single/global_best.py
bradahoward: It appears that using `**` outside of a function call throws an `IndexError`.  The Python documentation around this uses them without the decorators in the outside of the function call.
https://docs.python.org/dev/tutorial/controlflow.html#more-on-defining-functions
Thoughts?

pyswarms/single/global_best.py
bradahoward: Sure! I fixed the optimize prototypes. The only goofy thing there is the *args had to be listed before the other key word arguments i.e. ` def fun(foo, bar, *args, baz=3, **kwargs)`

ljvmiranda921:
Thanks for this @bradahoward! I'll go over this PR hopefully this weekend, just need to finish this week, etc. Noted on your questions, I will try some variants on my end and let's see what works.

ljvmiranda921:
Hi @bradahoward , 

I'm thinking of just using `**kwags` instead. Adding positional arguments may introduce nasty errors, and whenever we parametrize our objective function, we often add named arguments, right? Any thoughts?

```
# Sample usage
def f(x, a, b):
    """An objective function with two params a and b"""
    j = (x**2.0).sum(axis=1)
    return (a * j + b)

optimizer = GlobalBestPSO(n_particles=10, dimensions=2, options)
optimizer.optimize(f, iters=100, a=5, b=6)
```

No need for `*args`, might make things confusing. Let's just have the user explicitly define the parameter arguments of his/her custom-objective function as keyword args.

pyswarms/single/global_best.py
ljvmiranda921: Whoops. Yes, you are correct, just remove `**` and it should print. One last Python 3.X thing,  perhaps a `.format()` is better?

```python
name = 'ljvmiranda921'
print("Hi my name is: {}".format(name))
```

bradahoward:
hi @ljvmiranda921  -- thanks for the quick feedback!

If it were me, i would only use `**kwargs`.  I like the explicit definition of the parameters.  I will change this up today.

For the `.format()` on the logging i followed what you had in the `print_step` `cli_print` call .  I can change that as well.

ljvmiranda921:
Hi @bradahoward ,

Awesome! Everything looks good to me (LGTM) now. You're correct, I think explicitly naming the arguments via `**kwargs` is much better. If you wish to change the string formats in the print functions, it's fine, just open up another PR if you want. 

Anyway, super thankful for your help and contribution, I appreciate your help and you're very easy to work with!

PR #147, opened 2018-06-26
Add choice of topology and tests for the pyramid topology

whzup:
I added a `topology` attribute and the default values for the `single` optimizers (global and local). Also, I expanded the test files for the global search as well as the `conftest` file. 

See also: #142


ljvmiranda921:
Hi @whzup! Thanks a lot!

Before we go over this, I recommend that you: 

1. First update the development branch
2. Make a new branch from it, and 
3. Do a PR with that new branch.

As you can see, your previous commits still showed up. This might mess up our git history though. 

It would also be nice if we can only focus on one issue, in this case, solving the weird behavior in the Pyramid topology class. We can do the new feature (i.e., adding the topology parameter) next time. 

I’m not sure if the instructions above will entail the deletion of your fork and re-forking this repo again. Perhaps there is an easier option to just merge ljvmiranda921/pyswarms:development **into** whzup/pyswarms:development to fulfill step 1. 

Thanks a lot and I am looking forward to your PR!

ljvmiranda921:
Oops I just read your comment in the past PR. Wait a minute...

ljvmiranda921:
Hi @whzup!

Now I understand what you are trying to do. GlobalBestPSO should always use the Star topology (the same way the LocalBestPSO should always use the Ring topology). This is with respect to the literature we are basing PSO algorithms upon (Kennedy and Eberhart, 1995). In a sense, the topologies for GlobalBest and Localbest are already set in stone. 

Are the errors inherent in the Pyramid topology itself, or is it because of this new topology parameter?

Now, maybe we can try making a GeneralOptimizer (apart from GlobalBest and LocalBest) that takes a topology parameter. What do you think?

whzup:
Oh yes, I see what you mean. I might have been a bit rashing on this feature. I'll delete this branch again and start over so I don't have this commit mess again.

About the errors, they might only be a problem of the data types in the pyramid class, because one cannot create an array of integers with subarrays of different length in `numpy`.

I like the idea of a GeneralOptimizer! I'll open an issue with some thoughts about it.

PR #151, opened 2018-07-02
Create GeneralOptimizer class and fix index error in Pyramid class

whzup:
Created a `GeneralOptimizer` class for the implementation of new topologies. Updated the documentation and created a new test file for the new class. With the new test file, it became possible to test the `Pyramid` class (topology) more easily.
Fixed an index error that occurred in the `Pyramid` topology because of a type mismatch.

Resolves: #148

### DIFF
--- a/docs/api/pyswarms.single.rst
+++ b/docs/api/pyswarms.single.rst
@@ -22,3 +22,13 @@ pyswarms.single.local_best module
    :show-inheritance:
    :private-members:
    :special-members: __init__
+
+pyswarms.single.general_optimizer module
+---------------------------------
+
+.. automodule:: pyswarms.single.general_optimizer
+   :members:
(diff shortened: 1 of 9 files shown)


whzup:
Should I include a citation to [this paper](http://pubs.cs.uct.ac.za/archive/00000477/01/ParticleSwarm.pdf) in the `Pyramid` class doc string before you merge it? It might give us some scientific "back-up" for the implementation of the Delaunay triangulation.

ljvmiranda921:
Hi @whzup , I'll merge this one first. I will open up a new issue regarding documentation fixes, including the one you mentioned in #152 so that we can start preparing for v.0.3.0

ljvmiranda921:
Merged! Thanks for this @whzup ! You're awesome!

PR #155, opened 2018-07-05
Add Random topology class

whzup:
Added a new topology with random neighbours. Added documentation and
a test file for it and reworked the documentation of the
GeneralOptimizer class to incorporate all available topologies.
Simplified the fixture function for the GeneralOptimizer class.
Cited the relevant paper for the algorithm implemented.

I implemented the algorithm to compute the neighbours in a separate, private method inside the `Random` class because I thought it's cleaner that way. We need such a complicated algorithm because it's important that the "graph" we create when assigning neighbours is fully connected.

### DIFF
--- a/pyswarms/backend/topology/random.py
+++ b/pyswarms/backend/topology/random.py
@@ -8,6 +8,7 @@ This class implements a random topology. All particles are connected in a random
 
 # Import from stdlib
 import logging
+import itertools
 
 # Import modules
 import numpy as np
@@ -45,6 +46,7 @@ class Random(Topology):
         k : int
             number of neighbors to be considered. Must be a
             positive integer less than :code:`n_particles-1`
+
         Returns
         -------
         numpy.ndarray
@@ -138,7 +140,27 @@ class Random(Topology):
 
         This method computes the adjacency matrix of the topology using
         the randomized algorithm proposed in [TSWJ2013]. The resulting
-        topology is a connected graph.
+        topology is a connected graph. This is achieved by creating three
+        matrices:
+
+            * adj_matrix :  The adjacency matrix of the generated graph.
+                            It's initialized as an identity matrix to
+                            make sure that every particle has itself as
+                            a neighbour. This matrix is the return
+                            value of the method.
+            * neighbor_matrix : The matrix of randomly generated neighbors.
+                                This matrix is a matrix of shape
+                                :code:`(swarm.n_particles, k)`:
+                                with randomly generated elements. It's used
+                                to create connections in the adj_matrix.
+            * dist_matrix : The distance matrix computed with Dijkstra's
+                            algorithm. It is used to determine where the
+                            graph needs edges to change it to a connected
+                            graph.
+
+        .. note::   If the graph isn't connected, it is possible that the
+                    PSO algorithm does not find the best position within
+                    the swarm.
 
         Parameters
         ----------
@@ -153,9 +175,10 @@ class Random(Topology):
         numpy.ndarray
             Adjacency matrix of the topology
         """
+
         adj_matrix = np.identity(swarm.n_particles, dtype=int)
-        # Generate auxiliary matrix
-        aux_matrix = np.array(
+
+        neighbor_matrix = np.array(
             [np.random.choice(
                 # Exclude i from the array
                 np.setdiff1d(
@@ -163,17 +186,15 @@ class Random(Topology):
                 ), k, replace=False
             ) for i in range(swarm.n_particles)])
 
-        # Set elements to one
-        adj_matrix[np.arange(swarm.n_particles).reshape(swarm.n_particles, 1), aux_matrix] = 1
-        adj_matrix[aux_matrix, np.arange(swarm.n_particles).reshape(swarm.n_particles, 1)] = 1
+        # Set random elements to one using the neighbor matrix
+        adj_matrix[np.arange(swarm.n_particles).reshape(swarm.n_particles, 1), neighbor_matrix] = 1
+        adj_matrix[neighbor_matrix, np.arange(swarm.n_particles).reshape(swarm.n_particles, 1)] = 1
 
-        # Compute distance matrix
         dist_matrix = dijkstra(adj_matrix, directed=False, return_predecessors=False, unweighted=True)
 
-        # Generate connected graph
+        # Generate connected graph.
         while connected_components(adj_matrix, directed=False, return_labels=False) != 1:
-            for i in range(swarm.n_particles):
-                for j in range(swarm.n_particles):
+            for i, j in itertools.product(range(swarm.n_particles), repeat=2):
                     if dist_matrix[i][j] == 0:
                         adj_matrix[i][j] = 1
 
--- a/pyswarms/backend/topology/random.py
+++ b/pyswarms/backend/topology/random.py
@@ -1,0 +1,180 @@
+# -*- coding: utf-8 -*-
+
+"""
+A Random Network Topology
+
+This class implements a random topology. All particles are connected in a random fashion.
+"""
+
+# Import from stdlib
+import logging
+
+# Import modules
+import numpy as np
+from scipy.sparse.csgraph import connected_components, dijkstra
+
+# Import from package
+from ..import operators as ops
+from .base import Topology
+
+# Create a logger
+logger = logging.getLogger(__name__)
+
+
+class Random(Topology):
+    def __init__(self):
+        super(Random, self).__init__()
+
+    def compute_gbest(self, swarm, k):
+        """Update the global best using a random neighborhood approach
+
+        This uses random class from :code:`numpy` to give every particle k
+        randomly distributed, non-equal neighbors. The resulting topology
+        is a connected graph. The algorithm to obtain the neighbors was adapted
+        from [TSWJ2013].
+
+        [TSWJ2013] Qingjian Ni and Jianming Deng, “A New Logistic Dynamic
+        Particle Swarm Optimization Algorithm Based on Random Topology,”
+        The Scientific World Journal, vol. 2013, Article ID 409167, 8 pages, 2013.
+        https://doi.org/10.1155/2013/409167.
+
+        Parameters
+        ----------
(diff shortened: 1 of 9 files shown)


whzup:
@ljvmiranda921 the static/dynamic feature is already in the pipeline and almost ready for PR 😄. Just gonna wait for you to approve this one so I can work from a clean branch.
Please don't feel stressed by my many PRs and take as much time as needed to review them!

ljvmiranda921:
> Please don't feel stressed by my many PRs and take as much time as needed to review them!

Haha! No problem. Will probably review them next week, @whzup , after my final defense (wish me luck)!

whzup:
Hahaha, I thought I'm going to push out some PRs before I go work 😄. Sure do! Wish you success! Is the thesis going to be available online? Might take a glance at it if so 😋.

ljvmiranda921:
> I thought I'm going to push out some PRs before I go work 
Sure @whzup ! Just push the changes and I'll take care of this next week.
> Is the thesis going to be available online? Might take a glance at it if so.
Of course! The link can be found [here](https://ljvmiranda921.github.io/research/)

whzup:
Just updated the documentation for the Random topology to fit the changes I made in the other topologies 👍.

pyswarms/backend/topology/random.py
```diff
@@ -0,0 +1,180 @@
+# -*- coding: utf-8 -*-
+
+"""
+A Random Network Topology
+
+This class implements a random topology. All particles are connected in a random fashion.
```
ljvmiranda921: It would be nice if we add a short explanation on why `dijkstra` is used and how it relates to the `Random` topology. :smile:

pyswarms/backend/topology/random.py
```diff
@@ -0,0 +1,180 @@
...
+        The Scientific World Journal, vol. 2013, Article ID 409167, 8 pages, 2013.
+        https://doi.org/10.1155/2013/409167.
+
+        Parameters
+        ----------
+        swarm : pyswarms.backend.swarms.Swarm
+            a Swarm instance
+        k : int
+            number of neighbors to be considered. Must be a
+            positive integer less than :code:`n_particles-1`
```
ljvmiranda921: A bit nitpicky but an empty line before `Returns` will make everything consistent :+1:

pyswarms/backend/topology/random.py
```diff
@@ -0,0 +1,180 @@
...
+        # Compute distance matrix
+        dist_matrix = dijkstra(adj_matrix, directed=False, return_predecessors=False, unweighted=True)
+
+        # Generate connected graph
+        while connected_components(adj_matrix, directed=False, return_labels=False) != 1:
+            for i in range(swarm.n_particles):
+                for j in range(swarm.n_particles):
+                    if dist_matrix[i][j] == 0:
+                        adj_matrix[i][j] = 1
+
```
ljvmiranda921: Wanna try `itertools.product` for this nested for-loop? Here's the [documentation](https://docs.python.org/3/library/itertools.html#itertools.product). Sample usage:

```python
import itertools

for i, j in itertools.product(range(swarm.n_particles), repeat=2):
    # Not sure if this ternary operator works, if not, just use the normal one
    adj_matrix[i][j] = 1 if dist_matrix[i][j] == 0

```

ljvmiranda921: From my understanding, the elements in  `adj_matrix` are initially set to `1`. Then we have a for-loop that traverses the `dist_matrix` and updates the `adj_matrix` to `1` if the corresponding element in the `dist_matrix` is equal to `0`. 

Does this change anything? Perhaps it's `connected_components` that makes the difference but maybe you can explain it more clearly? :+1:

pyswarms/backend/topology/random.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        return ops.compute_position(swarm, bounds)
+
+    def __compute_neighbors(self, swarm, k):
+        """Helper method to compute the adjacency matrix of the topology
+
+        This method computes the adjacency matrix of the topology using
+        the randomized algorithm proposed in [TSWJ2013]. The resulting
+        topology is a connected graph.
+
```
ljvmiranda921: I think we can add a short explanation between the `adj_matrix`, `aux_matrix`, and `dist_matrix` here in order to guide other developers who may someday work with this code. :+1:

tests/backend/topology/test_random.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+# Import modules
+import pytest
+import numpy as np
+
+# Import from package
+from pyswarms.backend.topology import Random
+
```
ljvmiranda921: Please add a test for `__compute_neighbors` :+1: 
This seems to be an integral part of the topology and it would be nice if we can make sure that it will always work as expected. :100: 

You can construct a "smaller" swarm in `conftest.py` to make it more manageable.

tests/optimizers/conftest.py
```diff
@@ -11,61 +11,35 @@
...
+@pytest.fixture(scope="module",
+                params=[
+                    Star(),
+                    Ring(),
+                    Pyramid(),
+                    Random()
+                ])
+def general_opt_reset(top):
+    """Returns a GeneralOptimizerPSO instance that has been run and reset to check
     default value"""
```
ljvmiranda921: Awesome! :1st_place_medal:

tests/optimizers/test_general_optimizer.py
```diff
@@ -52,6 +67,21 @@ def test_invalid_k_or_p_values(options):
         GeneralOptimizerPSO(5, 2, options, Ring())
```
ljvmiranda921: Great tests! To be honest, the `GeneralOptimizer` can make everything [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). The reason we will still keep the GlobalBest and LocalBest is for tradition&mdash; they're the commonly-known PSO algorithms, and having them exposed to a first-time user is good.

pyswarms/backend/topology/random.py
whzup: I was wondering about this part as well. Only when I tested the algorithm without the condition that you test it for connectivity it became clear to me. If you don't have a connected graph it is quite possible that you can't extract the best position from the swarm. It might not be a problem with larger swarm sizes but in smaller ones, it occurred very often. I think this happens because the `compute_gbest()` method can't handle multiple unconnected swarm clusters. Just wanted to calrify this because it's hard to stuff all this information in a 3 line comment 😄. I'll go ahead and change these things in the evening 👍.

tests/backend/topology/test_random.py
whzup: What do you think is the best way to test it? The elements of the output matrix are random so we can't check if it returns the "right" elements. The only things we could test are the shape and the symmetry. We could use small swarms but then it would just be lucky if we get the right answers or not.

whzup: Implemented a test for the shape and the symmetry, I hope it is sufficient for you 🙈.

ljvmiranda921:
Thanks a lot @whzup ! This PR looks good and nicely-structured. 

> Thanks for the good inputs! Surely went well the thesis looks pretty good +1. 

Thank you so much! I appreciate that you liked it! 

> Done with all changes except the testing for the compute_neighbors() method. Not sure how to test it, TBH

So my idea is that you contrive a small swarm from the `Swarm` class and add it to `conftest.py` (or you can use the existing `swarm` fixture), then, try to pass it to `__compute_neighbors` and see the return value. That will then be your expected value.

After that, we write a test that checks if it will always give us the expected value whenever the test suite is run. If there is stochasticity (randomness) involved, you can set the random seed (not sure if you will set numpy's random seed or the random library's seed) so that it will always return the same thing. :+1:

whzup:
Done! Hope you like it 😄.

pyswarms/backend/topology/random.py
```diff
@@ -0,0 +1,193 @@
...
+        # Compute the distance matrix using Dijkstra's algorithm to
+        # later find unconnected pieces of the topology.
+        dist_matrix = dijkstra(adj_matrix, directed=False, return_predecessors=False, unweighted=True)
+
+        # Generate connected graph. If the graph isn't connected,
+        # it is possible that the PSO algorithm does not find the
+        # best position within the swarm. If the element (i,j) of the
+        # distance matrix is 0, that means that there is no path
+        # from node i to node j, thus there are still unconnected
+        # pieces of the graph (i.e. topology)
```
ljvmiranda921: Hi @whzup , one last thing regarding documentation. I'm not sure if you've seen my earlier comment but I suggested to put this long text on the docstring itself. You can just put a `# Generate connected graph` on the docstring. The idea is to make the docstring a bit longer while the in-line comments shorter (less clutter) and can be viewed by the developer using `help()` without resorting to look at the source code.

tests/backend/topology/test_random.py
ljvmiranda921: I really wonder why my earlier reviews didn't show up :confused: So you can probably try to set a random seed (not sure if we set the numpy random seed or the standard library's `random` seed) to some value so that it will always return the correct values whenever the `__compute_neighbors()` is called.

ljvmiranda921: So yes, just check if setting a random seed will work so that we always have an expected behaviour

pyswarms/backend/topology/random.py
```diff
@@ -0,0 +1,201 @@
...
+                            value of the method.
+            * neighbor_matrix : The matrix of randomly generated neighbors.
+                                This matrix is a matrix of shape
+                                :code:`(swarm.n_particles, k)`:
+                                with randomly generated elements. It's used
+                                to create connections in the adj_matrix.
+            * dist_matrix : The distance matrix computed with Dijkstra's
+                            algorithm. It is used to determine where the
+                            graph needs edges to change it to a connected
+                            graph.
```
ljvmiranda921: Awesome! This is good @whzup !

Issue #167, opened 2018-07-17
Fix plotter module 


jayspeidell:
* PySwarms version: 0.20
* Python version: 3.6
* Operating System: Linux

### Description

1. The designer class in formatters.py is used for 3D graphing, but the default label list attribute only includes two axes. This causes an index out of range error. 

2. The delta attribute of the Mesher class in formatters.py is 0.001. This is too detailed and can lead to mesher objects that are large enough to cause memory errors on most computers. 0.1 will give the same level of detail with much less memory usage. 

3. The colormap attribute is hardcoded, so there's no way for the user to choose their own. 

### What I Did

I've already made updates to the code addressing these issues and will make a pull request soon.

ljvmiranda921:
Conversation can be found in #165

PR #172, opened 2018-07-18
Fix plotter module (#167)

jayspeidell:
## Description

Bug fix: Added 'z-axis' to formatters/Designer attribute 'labels' to prevent index out of range error when plotting in 3D. Also added a third dimension to 'limits' for the same reason. 

New feature:  I added a colormaps property to the Designer class in formatters.py to allow the user to choose whichever one they want. It uses the same style as other properties of the class, and the module imports matplotlib.cm to validate both types of colormaps in matplotlib. I edited plotters/plot_countour to initialize a Designer object with a default colormap if none is provided. I've tested with every category of colormap as well as custom colormaps and verified that it is working as intended.

Default parameter change: Mesher.delta from 0.001 to 0.1. This makes it safe to run many of the functions on an average computer. With 0.001, the mesher object grows very quickly. 

## Related Issue

Fix plotter module (#167)

## Motivation and Context

Adding a colormap attribute to Designer made easier to verify the functions. I implemented because it enabled me to use discreet colormaps. 

## How Has This Been Tested?

Then I used flask8 and black. I restricted flask8 and black testing to the files that I edited in each commit. I also ran py.test. 

I tested Designer on all types of matplotlib colormaps. I didn't add a test to do this automatically, but it will return a TypeError if it's not a valid colormap. Does it need a test? 

## Screenshots (if appropriate):
sphere_example

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] All new and existing tests passed.

### DIFF
--- a/pyswarms/utils/plotters/formatters.py
+++ b/pyswarms/utils/plotters/formatters.py
@@ -10,6 +10,7 @@ This module implements helpful classes to format your plots or create meshes.
 import numpy as np
 from attr import attrs, attrib
 from attr.validators import instance_of
+from matplotlib import cm
 
 
 @attrs
@@ -57,10 +58,15 @@ class Designer(object):
     )
     legend = attrib(validator=instance_of(str), default="Cost")
     label = attrib(
-        validator=instance_of((str, list, tuple)), default=["x-axis", "y-axis"]
+        validator=instance_of((str, list, tuple)),
+        default=["x-axis", "y-axis", "z-axis"],
     )
     limits = attrib(
         validator=instance_of((list, tuple)), default=[(-1, 1), (-1, 1)]
+    )
+    colormap = attrib(
+        validator=instance_of((type(cm.viridis), type(cm.binary))),
+        default=cm.viridis,
     )
 
 
--- a/pyswarms/utils/plotters/formatters.py
+++ b/pyswarms/utils/plotters/formatters.py
@@ -10,7 +10,7 @@ This module implements helpful classes to format your plots or create meshes.
 import numpy as np
 from attr import attrs, attrib
 from attr.validators import instance_of
-from matplotlib import cm
+from matplotlib import cm, colors
 
 
 @attrs
@@ -65,7 +65,7 @@ class Designer(object):
         validator=instance_of((list, tuple)), default=[(-1, 1), (-1, 1), (-1, 1)]
     )
     colormap = attrib(
-        validator=instance_of((type(cm.viridis), type(cm.binary))),
+        validator=instance_of(colors.Colormap),
         default=cm.viridis,
     )
 
--- a/pyswarms/utils/plotters/formatters.py
+++ b/pyswarms/utils/plotters/formatters.py
@@ -62,7 +62,7 @@ class Designer(object):
         default=["x-axis", "y-axis", "z-axis"],
     )
     limits = attrib(
-        validator=instance_of((list, tuple)), default=[(-1, 1), (-1, 1)]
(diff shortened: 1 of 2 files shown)


ljvmiranda921:
> Bug fix: Added 'z-axis' to formatters/Designer attribute 'labels' to prevent index out of range error when plotting in 3D. Also added a third dimension to 'limits' for the same reason.

Understood.

> New feature: I added a colormaps property to the Designer class in formatters.py to allow the user to choose whichever one they want.

This is good! I guess it's better to add flexibility with respect to colormaps. :+1: Let me check your tests later.

> Default parameter change: Mesher.delta from 0.001 to 0.1. This makes it safe to run many of the functions on an average computer. With 0.001, the mesher object grows very quickly.

Copy that.

Let me do the review in a while :+1:

pyswarms/utils/plotters/formatters.py
```diff
@@ -10,6 +10,7 @@
 import numpy as np
 from attr import attrs, attrib
 from attr.validators import instance_of
+from matplotlib import cm
```
ljvmiranda921: I think the best way to validate the colormaps is to check if they are instances of the `matplotlib.colors.Colormap` class. Both `ListedColormap` (the type of `cm.viridis`) and `LinearSegmentedColormap` (the type of `cm.binary`) are subclasses of `Colormap`.

So my suggestion is:

```python
from matplotlib import cm, colors

class Designer(object):
    colormap = attrib(
        validator=instance_of(colors.Colormap),
        default=cm.viridis,
     )
```

Can you confirm if this works? :smile: 
If this works, I think there's no need to write tests for all colormaps since we have a validator that gives out an error whenever we enter a non-`Colormap`

pyswarms/utils/plotters/formatters.py
```diff
@@ -57,10 +58,15 @@ class Designer(object):
...
+        default=["x-axis", "y-axis", "z-axis"],
     )
     limits = attrib(
-        validator=instance_of((list, tuple)), default=[(-1, 1), (-1, 1)]
+        validator=instance_of((list, tuple)), default=[(-1, 1), (-1, 1), (-1, 1)]
+    )
+    colormap = attrib(
+        validator=instance_of((type(cm.viridis), type(cm.binary))),
+        default=cm.viridis,
     )
```
ljvmiranda921: This line of code is related to my comment above. I think it's better to check against `matplotlib.colors.Colormap` because it is the base class for whatever `type(cm.viridis)` and `type(cm.binary)` is

pyswarms/utils/plotters/formatters.py
jayspeidell: That makes sense! I had been trying to validate colors.Colormap with type(), but that's redundant with attrib and messes things up.

ljvmiranda921:
LGTM! Will merge in a while :+1:

ljvmiranda921:
Merged. Thank you so much @jayspeidell ! I appreciate this contribution! You're correct, I realized that we should decrease the resolution in the mesh to accommodate other machines. I also liked the fact that you opened-up flexibility for the users in choosing their colormaps. Good job! :tada: 

Thanks a lot for this wonderful PR. I hope you enjoyed despite the minor hassle of separating PRs the last time hehe :100: Aaron will take over the review of your other PR, we hope to see you again!

jayspeidell:
@ljvmiranda921 Awesome, thanks! I appreciate the opportunity to learn the process.

PR #175, opened 2018-07-19
Fix bug in setup.py

### LINKED ISSUE #174: setup.py does throws a UnicodeDecodeError
**Describe the bug**
The `setup.py` file throws a `UnicodeDecodeError`:
```
Traceback (most recent call last):
  File "setup.py", line 9, in <module>
    readme = readme_file.read()
  File "Python\Python37-32\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 11427: character maps to <undefined>
```

**To Reproduce**
Run the `setup.py` file to install `pyswarms`:

```
python setup.py install
```

**Expected behaviour**
No `UnicodeDecodeError` 😄.

**Environment (please complete the following information):**
 - OS: Windows
 - PySwarms Version v0.2.1
 - Python Version 3.7.0

**Additional context**
Not a big deal there is a [StackOverflow question](https://stackoverflow.com/questions/9233027/unicodedecodeerror-charmap-codec-cant-decode-byte-x-in-position-y-character) with the solution to this problem. PR is coming.

whzup:
## Description
Fixed a bug in the setup.py file where it would throw an error because the encoding format was not right. This is done by adding the encoding parameter to the open() function.

## Related Issue

Resolves #174 

## Motivation and Context

There might be users that cannot install the package using the `setup.py` file because of that.

## How Has This Been Tested?

I ran the `setup.py` file.

## Types of changes

- [X] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)

## Checklist:

- [X] All new and existing tests passed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,10 +5,10 @@
 
 from setuptools import setup, find_packages
 
-with open('README.md') as readme_file:
+with open('README.md', encoding="utf8") as readme_file:
     readme = readme_file.read()
 
-with open('HISTORY.rst') as history_file:
+with open('HISTORY.rst', encoding="utf8") as history_file:
     history = history_file.read()
 
 requirements = [


ljvmiranda921:
While you’re at it, maybe we can add a Semaphore CI to test on Windows. Windows is terrible with UTF8 and some CRLF, etc. Might also be the cause of the problem

whzup:
According to the [docs](https://docs.python.org/3/library/codecs.html#encodings-and-unicode):

>There’s another group of encodings (the so called charmap encodings) that choose a different subset of all Unicode code points and how these code points are mapped to the bytes 0x0–0xff. To see how this is done simply open e.g. encodings/cp1252.py (which is an encoding that is used primarily on Windows). There’s a string constant with 256 characters that shows you which character is mapped to which byte value.

So Python might just take `cp1252` as default decoding on Windows.
BTW, it also happens when I use Python 3.6.

ljvmiranda921:
Copy that. Let's add continuous-integration on Windows on v.0.4.0

PR #185, opened 2018-07-29
Ignore PyYAML dependency

ljvmiranda921:
I guess it's better to have pyup ignore the PyYAML dependency. We're only using yaml to configure our logging files so it's not really that relevant as compared to numpy or matplotlib


ljvmiranda921:
Any thoughts, @whzup ?

PR #186, opened 2018-07-29
Fix pyyaml dependency

ljvmiranda921:
It turns out that pyup and pyyaml has anomalous versions and
has become buggy. We fix this dependency by having pyup ignore
such changes. PyYAML is only used for our logging configs, so
this will not be harmful.

Signed-off-by: Lester James V. Miranda <ljvmiranda@gmail.com>


ljvmiranda921:
I'll override this for a while, @whzup  and @SioKCronin , just to have a quick fix for this bottleneck that's causing us a lot of PR errors.

Issue #183, opened 2018-07-28
`ValueError` when running the GeneralOptimizer with a `Pyramid` topology


whzup:
**Describe the bug**
The `Pyramid` topology throws the following error when ran in the `GeneralOptimizerPSO` (only in some cases):
```python
\AppData\Local\Programs\Python\Python37-32\lib\site-packages\src\pyswarms\pyswarms\backend\topology\pyramid.py", line 75, in <listcomp>
    [swarm.pbest_cost[self.neighbor_idx[i]].argmin() for i in range(len(self.neighbor_idx))]
ValueError: attempt to get argmin of an empty sequence
```

**To Reproduce**
Steps to reproduce the behaviour:

- Run this script in the current `development` version:
```python
import pyswarms as ps
import numpy as np
from pyswarms.utils.functions import single_obj as fx
from pyswarms.backend.topology import Random, Pyramid, Ring, Star, VonNeumann

# Set-up hyperparameters
options = {"c1": 2, "c2": 2, "w": 0.5, "k": 4, "p": 2, "r":1}
iters = 1000
dim = 5
bounds = (np.repeat(-5.12, dim), np.repeat(5.12, dim))
topology = Pyramid(static=False)
optimizer = ps.single.GeneralOptimizerPSO(n_particles=10, dimensions=dim, options=options, bounds=bounds, topology=topology)
best_cost, best_pos = optimizer.optimize(fx.sphere_func, iters=iters, verbose=0, print_step=1)
```

**Environment:**
 - OS: Windows
 - PySwarms Version development
 - Python Version 3.7.0

**Additional context and bug fix**
I investigated this bug and it turns out this might be a precision error in the `Delaunay` class which is used in the `Pyramid` topology to compute the neighbours. It seems that if two particles are too close to each other the triangulation isn't computed correctly. This leads to some particles not having any neighbours, that is, their array in the `neighbour_idx` attribute being empty. I looked up the [scipy documentation](https://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.spatial.Delaunay.html) which led me to the [qhull documentation](http://www.qhull.org/html/qdelaun.htm). I suggest that we change this line:

```python
pyramid = Delaunay(swarm.position)
```

to the following:

```python
pyramid = Delaunay(swarm.position, qhull_options="QJ0.001 Qbb Qc Qx")
```
What the options do:

[...]

PR #180, opened 2018-07-25
Prepare v.0.3.0 release

ljvmiranda921:
## Tasklist (Deadline: August 10, 2018)
- [x] Run `black` and `flake8` for a clean format (drop)
- [x] Tests and minor changes for topologies (c/o @whzup )
- [x] Make a PyPI friendly README by following this [link](https://packaging.python.org/guides/making-a-pypi-friendly-readme/)
- [x] Boundary conditions #150 (c/o @jolayfield) *Moved to v.0.4.0 or on dev branch*
- [x] Resolve issue with PyYaml, you can find a related issue in yaml/pyyaml#207
- [x] Update README (update authors and maintainers)
- [x] Merge all dependency updates in #162, #166, #173, and #178 
- [x] Resolve all merge conflicts
- [x] Update HISTORY.rst 
- [x] Write Release Notes
- [x] Bump version and update setup.py
- [x] **Release:** merge to `master`, make a Release, and upload to PyPI

*Please write the changelog below for our release notes. It would be nice to follow [this](https://github.com/ljvmiranda921/pyswarms/releases/tag/v.0.2.0) link.*

---
# Release v.0.3.0

We're proud to present the release of PySwarms version 0.3.0! Coinciding with this, we would like to welcome Aaron Moser (@whzup) as one of the project's maintainers! v.0.3.0 includes new topologies, a `static` option to configure a particle's neighbor/s, and a revamped `plotters` module. We would like to thank our contributors for helping us with this release.

## Release notes

- **NEW:** More basic particle topologies in the `pyswarms.backend` module - #142, #151, #155, #177 
- **NEW:** Ability to make topologies static or dynamic - #164 
- **NEW:** A `GeneralOptimizerPSO` class. The `GeneralOptimizerPSO` class has an additional attribute for the topology used in the optimization - #151
- **NEW:** A `plotters` module for swarm visualization. The `environments` module is now deprecated - #135, #172 
- **FIX:** Bugfix for optimizations not returning the best cost - #176 
- **FIX:** Bugfix for `setup.py` not running on Windows - #175
- **IMPROVED:** Objective functions can now be parametrized. Helpful for your custom-objective functions - #144. Thanks, @bradahoward!
- **IMPROVED:** New single-objective functions - #168. Awesome work, @jayspeidell! 

## New Topologies and the `GeneralOptimizerPSO` Class

[...]

PR #189, opened 2018-07-29
Clean up changes for v0.3.0

whzup:
I cleaned up the code for the release of v0.3.0. The changes I made include:

- Set the `Star` topology to static and only assign the `neighbor_idx` attribute if it is not already assigned.
- Added the `VonNeumann `topology to the `GeneralOptimizer `with the necessary changes to the documentation and new exceptions for missing attributes.
- Added the `VonNeumann `topology to the `GeneralOptimizer `tests. The exceptions are tested and the topology fixture now also uses the `VonNeumann `class. Cleaned up the tests and added a static parametrization.
- Fixed a bug in the `Pyramid `topology where the optimize function of the `GeneralOptimizer `would output a `ValueError `because for some particles the neighbours were not computed. See also: #183 
- Changed the boundary for the connectivity of the particles in the `Pyramid `topology to be `swarm.dimensions + 1` instead of `5`. The reason for this change is that an `n`-dimensional simplex has `n+1`vertices.
- Ran `flake8`

See also: #183, #180 
Resolves: #183

### DIFF
--- a/pyswarms/single/general_optimizer.py
+++ b/pyswarms/single/general_optimizer.py
@@ -64,7 +64,7 @@ import numpy as np
 # Import from package
 from ..base import SwarmOptimizer
 from ..backend.operators import compute_pbest
-from ..backend.topology import Topology, Ring, Random
+from ..backend.topology import Topology, Ring, Random, VonNeumann
 from ..utils.console_utils import cli_print, end_report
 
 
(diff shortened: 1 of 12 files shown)


pyswarms/single/general_optimizer.py
```diff
@@ -174,21 +182,38 @@ def __init__(
...
+                    self.r <= 0
+                    or not 0
+                           <= VonNeumann.delannoy(self.swarm.dimensions, self.r)
+                           <= self.n_particles - 1
+                ):
+                    raise ValueError(
+                        "The range must be set such that the computed"
+                        "Delannoy number (number of neighbours) is"
+                        "between 0 and the no. of particles."
+                    )
```
ljvmiranda921: I'll accept this for now, but for v.0.4.0 we should make our if-statements shallower. In addition, it would be nice if we can create (for the next version) an `errors.py` module for our own custom exceptions. That would be nice. I'll put that in the pipeline.

PR #193, opened 2018-07-31
Added a Vagrantfile for easier collaboration

jdbohrman:
Hello, I've tested the build and it's passing on my system so I hope this helps with faster collaboration.


.vagrant/machines/default/virtualbox/vagrant_cwd
```diff
@@ -0,0 +1 @@
+/home/jdbohrman/Documents/pyswarms
```
ljvmiranda921: I'm wondering if the hardcoded paths are necessary.

.vagrant/machines/default/virtualbox/synced_folders
```diff
@@ -0,0 +1 @@
+{"virtualbox":{"/vagrant":{"guestpath":"/vagrant","hostpath":"/home/jdbohrman/Documents/pyswarms","disabled":false,"__vagrantfile":true}}}
```
ljvmiranda921: I'm wondering if the hardcoded paths are necessary.

.vagrant/machines/default/virtualbox/vagrant_cwd
jdbohrman: Hey @ljvmiranda921 great to be here! Here's the info I found on it. The documentation is slim but it appears that is the default insecure private key that Vagrant ships with. Here's a couple links: 

https://www.vagrantup.com/docs/vagrantfile/ssh_settings.html
https://github.com/hashicorp/vagrant/issues/8058

Also, I'm fairly sure you're absolutely correct, as running `vagrant up` would provision the config files and others.

ljvmiranda921:
Note to self: merge to `development` after v.0.3.0 release.

PR #218, opened 2018-08-13
Update base classes to inherit from ABC

ljvmiranda921:
Reference: #198 
There's really no big change aside from the ABC.
[More about Abstract Base Classes](https://pymotw.com/3/abc/)


ljvmiranda921:
> What exactly is this good for?

Turns out this is the standard way of implementing base classes, not by inheriting from an `object` class. It also unlocks other attributes and has stricter subclass checking for abstract methods (for `SwarmOptimizer`, it really ensures that `optimize()` is implemented)

PR #226, opened 2018-08-15
Add cost function decorator

whzup:
## Description

This decorator allows the creation of much simpler cost functions. Instead of writing a cost function that returns a shape of `(n_particles, 0)` it enables the usage of shorter and simpler cost functions that directly return the cost.

## Related Issue

Resolves: #228 

## Motivation and Context

At the moment, the way one has to write the cost function is very counterintuitive as it is necessary to write a function that returns an array of shape `(n_particles, )`. This decorator "vectorizes" a cost function which returns just the cost.
 
## How Has This Been Tested?

Wrote a new test file.

## Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [X] New feature (a non-breaking change which adds functionality)

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [x] All new and existing tests pass.


pyswarms/utils/decorators/decorators.py
```diff
@@ -0,0 +1,29 @@
...
+    """A decorator to mark the cost function
+
+    This decorator provides a way to simplify the cost function. The cost function
+    needs to resemble a strategy that can be applied to every particle in the swarm
+    to assign a cost. This means that the decorated cost function must use an array
+    of shape :code:`(dimensions, )` as an argument and return a :code:`float` or an
+    :code:`int`.
+    
+    Attributes
+    ==========
```
ljvmiranda921: I think this should be a second-level header, and `Parameters`:

```python
def cost(cost_func):
    """

    Parameters
    ----------
    """
```

Same with `Returns` :+1:

whzup: Yes, thanks for the catch 😄

pyswarms/utils/decorators/decorators.py
```diff
@@ -0,0 +1,29 @@
...
+    This decorator provides a way to simplify the cost function. The cost function
+    needs to resemble a strategy that can be applied to every particle in the swarm
+    to assign a cost. This means that the decorated cost function must use an array
+    of shape :code:`(dimensions, )` as an argument and return a :code:`float` or an
+    :code:`int`.
+    
+    Parameters
+    ----------
+    
+    cost_func : function
```
ljvmiranda921: This should be `callable` instead of `function`

pyswarms/utils/decorators/decorators.py
```diff
@@ -0,0 +1,29 @@
...
+        A function that can be used as cost function in the optimization (must return a :code:`float` or an :code:`int`)
```
ljvmiranda921: This line's a bit long (?) I don't think black formats docstrings, can `flake8` catch this line?  What does `flake8 decorators.py` say?

tests/utils/decorators/test_decorators.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @cost
+    def cost_func_w_decorator(x):
+        cost = objective_func(x)
+        return cost
+
+    w_o_decorator = cost_func_w_o_decorator(particles)
+    w_decorator = cost_func_w_decorator(particles)
+
+    assert isinstance(w_decorator, Number)
```
ljvmiranda921: Hmmm, I'm not sure why we need to test if the output is a `Number`. Maybe we should test the shape of the resulting array? Ensure that it's `(n_particles, )`?

pyswarms/utils/decorators/decorators.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Returns
+    -------
+
+    cost_dec : function
+        The vectorized output for all particles as defined by :code:`cost_func`
```
ljvmiranda921: This should be:

```python
"""
Returns
----------
callable
    My description here
"""
```

pyswarms/utils/decorators/decorators.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def cost(cost_func):
+    """A decorator to mark the cost function
+
+    This decorator provides a way to simplify the cost function. The cost function
+    needs to resemble a strategy that can be applied to every particle in the swarm
+    to assign a cost. This means that the decorated cost function must use an array
+    of shape :code:`(dimensions, )` as an argument and return a :code:`float` or an
+    :code:`int`.
```
ljvmiranda921: I think this can be phrased as:

> If you have defined a cost function that only acts upon a single particle, then you can use this decorator to have a vectorized computation for *n* particles. ...

Then give an example `code-block` maybe? I think good keywords here are: *wrapper*, *transforms*, etc. *Strategy* seems a bit odd (I think you got this from hypothesis haha but I think it's more of their local lingo).

tests/utils/decorators/test_decorators.py
```diff
@@ -0,0 +1,32 @@
...
+
+# Import from package
+from pyswarms import cost
+
+
+@pytest.mark.parametrize(
+    "objective_func",
+    [np.sum(), np.prod()]
+)
+def test_cost_decorator(objective_func, particles):
```
ljvmiranda921: Where do you get the particles? I suggest you can generate them via `np.random.uniform`...

tests/utils/decorators/test_decorators.py
```diff
@@ -0,0 +1,32 @@
...
+    def cost_func_w_o_decorator(x):
```
ljvmiranda921: It's ok to spell it out :laughing: `cost_func_without_decorator(x)` and `cost_fun_with_decorator`

ljvmiranda921:
Aside from my comments, I think this is great, @whzup ! This is an elegant and useful improvement to our current codebase! Very excited to try this out!

whzup:
@ljvmiranda921 Thanks for the comments hahaha. Would've been searching for ages to find a solution to these things 🙈

tests/utils/decorators/test_decorators.py
```diff
@@ -0,0 +1,32 @@
...
+import numpy as np
+from numbers import Number
+
+# Import from package
+from pyswarms import cost
+
+
+@pytest.mark.parametrize(
+    "objective_func",
+    [np.sum(), np.prod()]
```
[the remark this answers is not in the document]
whzup: I'd have to define a new function for this since `np.dot` takes two arguments. Shall I put it in the `conftest.py` file?

tests/utils/decorators/test_decorators.py
whzup: Do you know a way we can output the seed number with the tests without using `pytest -s`? In this way, we don't have fixed particle arrays and if it fails we can use the printed seed to debug it.

whzup:
Hmm, why do the fmt and the pyup commit show up here? Is this because I rebased the commits?

whzup:
Let's update the branch before we merge this one. I had some trouble with the rebasing of the commits.

tests/utils/decorators/test_decorators.py
```diff
@@ -0,0 +1,31 @@
+# Import modules
+import pytest
+import numpy as np
+import sys
```
ljvmiranda921: I wonder where you'll use this for... :man_astronaut:

whzup:
Ok, I guess we're ready for the merge. I think we can add change the representation of cost functions in another PR. What do you think?

ljvmiranda921:
Hi @whzup 

> I think we can add change the representation of cost functions in another PR. What do you think?

You mean the `single_obj` cost functions? I think there's no need to change them since they're all vectorized already.

> Ok, I guess we're ready for the merge. 

OK let me check first then I'll squash-merge this :+1:

pyswarms/__init__.py
```diff
@@ -16,5 +16,8 @@
 
 from .single import global_best, local_best, general_optimizer
 from .discrete import binary
+from .utils.decorators import *
+from . import utils
```
ljvmiranda921: Hmmm, I think the previous version was enough where `cost` is defined explicitly.
I think the `@constraint` decorator in the future is better loaded whenever the user needs it.

pyswarms/utils/decorators/decorators.py
```diff
@@ -0,0 +1,47 @@
+import numpy as np
+
+
+def cost(cost_func):
```
ljvmiranda921: :100: This is good already

tests/utils/decorators/test_decorators.py
```diff
@@ -0,0 +1,30 @@
...
+
+# Import from package
+from pyswarms import cost
+
+
+@pytest.mark.parametrize(
+    "objective_func",
+    [np.sum, np.prod]
+)
+def test_cost_decorator(objective_func, particles):
```
ljvmiranda921: The tests are good already :100:

ljvmiranda921:
> Although, what do you think about the numpy approach? Almost everything in their package can be accessed by using np.<name>

I thought about that before, that's why there's this presence of `import pyswarms as ps` that I "copied" from the `numpy` style. The only thing that's stopping me from doing this is that we'll have instances of this:

```python
import pyswarms as ps

ps.single.GlobalBestPSO # Good enough
ps.utils.functions.single_obj.sphere # Pretty long
ps.utils.decorators.decorators.cost # Kinda long too..
```

If it is possible to do `ps.sphere` or `ps.cost` but still respect the modules themselves, then maybe we can do something like that. We can load everything right away, but I'm just wary that we might have conflicting methods or something in the future.

That's the reason why I decided to have everything be loaded whenever needed. Atleast I know that users will always come for the high-level optimizers (and I'm sure the `cost` decorator too), but they won't really use all objective functions, and might clutter our namespace.

Right now, the design kinda follows that of Keras, where we have components separated into modules, and we load them whenever needed. The thing about `numpy` is that instead of components, you have operators which you will use regularly. It would then be a hassle if numpy does something similar to ours.

For now just explicitly declare `cost`, we can discuss this in the later `0.X` versions :+1:

ljvmiranda921:
> No, I mean all cost functions in our examples :smile:

Let's open-up an issue after I merge #227. We'd probably update some of our examples later + README (more animations haha)

whzup:
Done! :+1:

PR #230, opened 2018-08-17
[WIP] Port to reporter

ljvmiranda921:
Port module to use `Reporter`


ljvmiranda921:
Please use logger.exception for all raise calls so that the whole stack trace is included in the log. self.rep.exception()

Issue #237, opened 2018-08-26
Add boundary handlers for stuck particles


ljvmiranda921:
**Is your feature request related to a problem? Please describe.**
There are some stuck particles as reported in #234 and #150 . This is related to `compute_position` waiving position updates whenever a particle will potentially end up out-of-bounds. There are standard ways to avoid this problem. The reflective bounds is one option, but the related PR is currently stale, so it may be important to create a general framework for handling the boundary problem.

**Describe the solution you'd like**
Create a set of handlers, such as a `BoundaryHandler`. Usage example:

Usage outside of the class

```python
# Usage outside of the class
from pyswarms.backend import operators as ops
from pyswarms.backend.handlers import BoundaryHandler

bh = BoundaryHandler(strategy="reflective") # And other arguments here if it's parameterized
ops.compute_position(swarm, bounds, handler=bh) 
```

Inside `BoundaryHandler`

```python
# Inside BoundaryHandler
class BoundaryHandler(object):
    def __call__(self, position, bounds, *args, **kwargs): # parameterized
        # Some logic here depending on the type of handler, for example:
        # (might be better as a dictionary with a string as key and callable as
        # its value)
        if self.strategy == "reflective":
            new_positions = self.reflective(position, bounds) # parameterized

    def reflective(self, position, bounds):
        """Some docstring here"""
        pass
```

How it will work in the backend:

```python
if handler:
    new_position = handler(swarm.positions, bounds)
```

ljvmiranda921:
What do you think, @whzup

whzup:
I like it! :smile: Let's do something like that.
Although, do you really think it's necessary to pass them as a list? We can't apply multiple boundary handlers at the same time

ljvmiranda921:
> I like it! smile Let's do something like that.

Yes maybe we can work on the same PR branch? There's also the need to use handlers for Velocity (check your link in Gitter). Maybe we can split the work between us?

ljvmiranda921:
> I wonder if it might suffice to use a VelocityHandler. If the particles don't even leave the bounds there is no reason to fix the position thinking Maybe we can try it out like that?

I think the two handlers solve different problems. Our perspective now should shift from "how to prevent particles from going out-of-bounds" into "what happens when they're out of bounds?" 
[...]

PR #234, opened 2018-08-21
[WIP] Add Jupyter Notebook to visualise topologies

whzup:
## Description

This is a Jupyter notebook with animations of different topologies and some general information about the topologies themselves.

## Related Issue

#156 

## Motivation and Context

The goal of this PR is to visualise the differences between the topologies that have been implemented so far. Additionally, it provides another resource where one might look up how to use the `pyswarms` library. A Jupyter notebook is added where animations of different topologies are shown so one can compare the different neighbourhood properties visually.

## How Has This Been Tested?

**TBD**

## Screenshots (if appropriate):
**TBD**

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

**TBD**

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #150, opened 2018-06-29
Boundary Conditions


jolayfield:
High-dimensional searches with the current bounds conditions result in a lot of lost optimization steps where many particles do not move.  

Multiple solutions exist to deal with these issues. We want to implement reflective boundary conditions that will keep the particle within the bounds in a way that still samples the whole parameter space.

ljvmiranda921:
Ok, interesting. I’ve been reading about reflective boundary conditions and it seems a nice feature to have. Looking forward to your pull request! :) Don’t hesitate to ask questions if you’re stuck :)

jolayfield:
Thanks.  I would like to ask a few questions:

1) Is it possible/advisable for us to add a third option to the tuple for bounds so that it would look like  (lb, up, bound_type)?  Can we make these kinds of decisions as they seem best to us or is it better to ask first so that we don't conflict with other ongoing work?
  
2) I am not exactly sure how to run/make the test suite.  I tried to run the options found on the contributions page, but was not sure whether I implemented them properly.  Is it better to hash this out now or wait until we are ready to merge the code?  

3) Is this the appropriate place to ask these questions?

Thanks

ljvmiranda921:
Hi @jolayfield ! Any updates on this?

We're planning a v.0.3.0 release by the end of August. If this feature won't make it next month, we'll just insert this feature on v.0.4.0. There's no rush, what you just need to do is to *update your own master branch*. 

We're here to help you anytime!

jolayfield:
Thanks.  We just got a working implementation yesterday with my research student.  I think there is still some testing to do and we will need to cleanup the input but it is coming along.  We will followup when we are sure that things are working properly (hopefully in the next few weeks).

ljvmiranda921:
This has been a bit stale, closing this for now. I'll open this up again if there are already updates :+1:

Issue #232, opened 2018-08-21
Refactor if-else hierarchy in GeneralOpt


ljvmiranda921:
The if-else hierarchy we have works, [but looks convoluted & unmaintainable](https://github.com/ljvmiranda921/pyswarms/blob/development/pyswarms/single/general_optimizer.py#L250). Better refactor this with private functions or a `dict`.

Or maybe we should work on how the backend topologies were implemented. I'm thinking of using `*args` and `**kwargs` for this, still not sure.

Reference: https://stackoverflow.com/questions/14626279/inheritance-best-practice-args-kwargs-or-explicitly-specifying-parameters

PR #238, opened 2018-08-31
Add BoundaryHandler and VelocityHandler

whzup:
## Description
This PR implements the new `BoundaryHandler` and `VelocityHandler` classes that deal with the issue of particles that don't move in optimisations with boundary conditions. This is achieved by using different strategies that have been described in [Hel10](https://opus4.kobv.de/opus4-fau/frontdoor/index/index/year/2010/docId/1328) (P. 28). These strategies include:
+ Nearest bound: Reset particle to the nearest boundary
+ Random: Reset the particle to a random position within the bounds
+ Shrink: Shrink the velocity such that the particle touches the boundary instead of surpassing it
+ Intermediate: Reset the particle to a value between its former location and the bound
+ Resample: Calculate new velocities until the next particle position is within the bounds
+ Periodic: Tile the space with the same search space over and over

There is one more:
+ Reflective: Mirror the position of the particle at the surpassed boundary

I won't include the last one in this PR as it is probably provided by @jolayfield. After they've made their PR I'm going to embed their code in the `BoundaryHandler` class.
The strategies for the `VelocityHandler` include:
+ Unmodified: The velocity remains without modification
+ Adjust: The velocity is adjusted to be exactly the distance vector between the previous and the current position.
+ Invert: The velocity is inverted.
+ Zero: The velocities of particles that go out-of-bounds are set to zero.

## Related Issue
Resolves: #237
See also: #150, #234 

## Motivation and Context
In #234 we encountered a weird phenomenon when optimizing with boundary conditions. A lot of particles stopped moving while the optimization was still running. This is caused by the waiving of updating the position in the `compute_position()` function when it leaves the boundaries. To fix this issue we introduce a new class called `BoundaryHandler` as well as a `VelocityHandler` which are going to reposition the particle instead of not updating it.

## Types of changes
- [ ] Bug fix (a non-breaking change which fixes an issue)
- [X] New feature (a non-breaking change which adds functionality)

## Checklist:
- [X] My change requires a change to the documentation.
- [ ] All new and existing tests pass.

## TODO:
- [x] Add `BoundaryHandler`
- [ ] Write tests for the `BoundaryHandler` strategies
- [x] Integrate the `BoundaryHandler` into the `operators`
- [x] Add `VelocityHandler`
- [ ] Write tests for the `VelocityHandler` strategies
- [x] Integrate the `VelocityHandler` into the `operators`
- [ ] Clean up documentation
- [x] Add functions for the strategies in LaTeX
- [ ] Check doc build for any LaTeX errors
- [ ] Clean up PR


whzup:
I think we should put the Resample and the Shrink strategy into the `VelocityHandler` as they require, unlike the others, the velocity to fix the particles position.

pyswarms/backend/handlers.py
```diff
@@ -0,0 +1,163 @@
...
+        elif self.strategy == "reflective":
+            new_position = self.reflective()
+        elif self.strategy == "shrink":
+            new_position = self.shrink()
+        elif self.strategy == "random":
+            new_position = self.random()
+        elif self.strategy == "intermediate":
+            new_position = self.random()
+        elif self.strategy == "resample":
+            new_position = self.resample()
```
ljvmiranda921: I'm wondering if it's better to use a dictionary for pattern matching...

```python
x = {"nearest": self.nearest, "resample": self.resample}
x(position, bounds, *args, **kwargs)
```

pyswarms/backend/handlers.py
```diff
@@ -0,0 +1,163 @@
...
+        -------
+        numpy.ndarray
+            the adjusted positions of the swarm
+        """
+        # Assign new attributes
+        self.position = position
+        self.lower_bound, self.upper_bound = bounds
+        self.__out_of_bounds()
+
+        if self.strategy == "nearest":
```
ljvmiranda921: Maybe it's better to just pass `position`, `bounds`, `*args` and`**kwargs` to each of these methods, so that it's easier to test them later on.

pyswarms/backend/handlers.py
```diff
@@ -0,0 +1,163 @@
...
+
+        This method resets particles that exceed the bounds to the nearest
+        available bound. For every axis on which the coordiantes of the particle
+        surpasses the boundary conditions the coordinate is set to the respective
+        bound that it surpasses.
+        """
+        bool_greater = self.position > self.upper_bound
+        bool_lower = self.position < self.lower_bound
+        self.position = np.where(bool_greater, self.upper_bound, self.position)
+        self.position = np.where(bool_lower, self.lower_bound, self.position)
```
ljvmiranda921: I think it's better to just return `new_position` rather than updating `self.position`. Lesser side-effects I guess so that we can test them.

pyswarms/backend/handlers.py
```diff
@@ -0,0 +1,163 @@
...
+import logging
+
+import numpy as np
+
+from ..util.reporter import Reporter
+
+rep = Reporter(logger=logging.getLogger(__name__))
+
+
+class BoundaryHandler:
```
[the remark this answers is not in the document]
whzup: Sure, it's not necessary anymore, though. https://stackoverflow.com/questions/1238606/is-it-necessary-or-useful-to-inherit-from-pythons-object-in-python-3-x

pyswarms/backend/handlers.py
whzup: Wouldn't it be x["nearest"](position, bounds, *args, **kwargs) then?

pyswarms/backend/handlers.py
whzup: I see, I'm not quite sure what we're going to do with the args and kwargs, though? We either need the positions or the velocity not much more :confused:

pyswarms/backend/handlers.py
whzup: Understood, I'll see how I can handle that. :+1:

whzup:
I'm just wondering if we need the `VelocityHandler` if we already handle the position. Shrink and Resample use the velocity so if we handle both the velocity and the position and the same time we do things twice if not thrice. I don't see a difference between repairing particle positions after they moved outside the bounds and making the same reparations just before the particle exceeds the boundaries. The velocity handlings in the dissertation seem very similar to the position handlings we will implement.

pyswarms/backend/handlers.py
ljvmiranda921: Yup you are correct

pyswarms/backend/handlers.py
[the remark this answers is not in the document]
ljvmiranda921: Yup, I'm aware. It's an artifact of us support Python 2.X before

pyswarms/backend/handlers.py
ljvmiranda921: As a note, this is also what I'm planning to achieve in the `GeneralOptimizer` (#232). We want to remove all those `if-else`'s since it becomes a handful given different topologies requiring different sets of inputs. Maybe the `compute_*` methods should just maintain the same call as with the base class, and have all others be delegated as a `**kwargs` argument.

ljvmiranda921: Thus, if we only have this as our main call:
```python
try:
    new_position = self.strategies[strategy](position, bounds, *args, **kwargs)
except KeyError:
    rep.log("Something") # Logger error 
    raise
else:
    return new_position
```

everything becomes much easier

pyswarms/backend/handlers.py
ljvmiranda921: Sure, so instead of passing the result of `np.where` into `self.position` (updating `self.position`), just give it to `new_position` and then return it.

ljvmiranda921:
I think the reason why the author adds a set of velocity handlers is that we also want to update the velocity matrix that caused these particles to be out-of-bounds. If they are all *Unmodified*, there's a chance that they'll go out-of-bounds again. Sure, we can call `BoundaryHandler` again, but it's better to pull the particles back inside the specified limits

ljvmiranda921:
Some things I changed:
- I simply applied all the comments in the earlier discussion. Instead of using `if-else` in calling the methods, we just hash through a dictionary (`self.strategies`).
- The dictionary `self.strategies` is generated automatically. This removes the need to manually modify parts of the code whenever a new strategy is added. It looks into class members without `_` or `__` (dunder methods)
- Add `try-except-else` structure to `nearest`. I think it's nice if we apply the same structure to all of our strategies.

I am playing with this concept because I'm looking to implement the same design in the `GeneralOptimizer`. I just want to have one (1) API call where all subtleties or differences are handled in the topologies themselves.

whzup:
@ljvmiranda921, :smile: thanks for the help! Was planning to make the changes while travelling home today (4hrs of train). Although, I would've struggled with the dictionary work to be honest. It looks nice now!
I thought about the `VelocityHandler` and I had the idea that we could handle the position and the velocity simultaneously because you said that you were planning on handling the velocities of the infeasible particles anyways.
It would make the whole handling easier although it might make it harder to choose different velocity and position handlings (e.g. intermediate for the velocity handling and random for the position handling) which would be quite interesting to play around with. What do you think?

ljvmiranda921:
> thanks for the help! Was planning to make the changes while travelling home today (4hrs of train). Although, I would've struggled with the dictionary work to be honest. It looks nice now!

No problem! Glad you liked it!

> I thought about the VelocityHandler and I had the idea that we could handle the position and the velocity simultaneously because you said that you were planning on handling the velocities of the infeasible particles anyways.

> It would make the whole handling easier although it might make it harder to choose different velocity and position handlings (e.g. intermediate for the velocity handling and random for the position handling) which would be quite interesting to play around with. What do you think?

I think it's better if we separate "handling" such that `BoundaryHandler` only handles positions and then `VelocityHandler` handles velocities. I think we can apply your idea to our high-level optimizers. They can add a `bh` and/or a `vh` depending on what they want :+1: 

> The next steps would now be:

Yes you are correct. We can do `VelocityHandler` too in a separate PR, but let's finish this first (no pressure and take your time!). I'm going home to my country on the 18th so I will again be MIA by that week.

whzup:
Ok :+1: What means MIA though? :smile: Gonna start building up the test suite for the `Handlers` now. BTW, there is some syntax error for Python 3.4 in the Travis CI. It's something about a `**` I guess.

whzup:
@ljvmiranda921 What do you think is the best way to write the tests for the handling strategies? I thought about making fixtures with which we can test (1) whether the particles are inside the bounds after we apply the `BoundaryHandler` and (2) if the particles are where they should be after the respective strategy has been applied. (1) should not be a big problem as we can just see if all values are inside the boundaries but I think (2) might be a bit tricky. Especially for the more "complex" strategies, e.g. "shrink", it might be hard to find out the values where they are going to land to. Maybe there is some other property we can easily check to see if it works properly?

whzup:
@ljvmiranda921 Another thing: Do you think it's a good idea to pass the history to the `compute_position()` method? For the "intermediate" and "shrink" it might make sense to have this information available. Another possibility would be to use the velocity and look if the position is out of bounds after adding the velocity. But this would mean that we correct the position **before** the particle leaves the boundaries. The "random" and "nearest" strategies correct the position **after** the particle has left the boundaries. This might cause some problem and may be hard to maintain. Also, the "resample" strategy requires us to modify the velocity a lot of times so this already is a special snowflake and it might not be a good idea to introduces a lot of different types here.

ljvmiranda921:
Adding history on compute position doesn’t seem pure design-wise. How about as a keyword argument?

Maybe the boundary handler can return both position and velocity? Not really sure how to deal with that though. 

Can we just implement 2 or 3 of the easiest techniques for now?

whzup:
Hmm, how about we store the history inside the swarm itself so we could pass the swarm as a keyword argument (which might be necessary for some)? It's a "data class" anyways.

> Can we just implement 2 or 3 of the easiest techniques for now?

I have implemented the "resample" strategy now so we have 3. Shall I go on and work on the tests and the integration in the `compute_something()` functions then?

ljvmiranda921:
I'm not sure why I cannot push into your PR anymore: https://github.com/whzup/pyswarms/pull/1

pyswarms/backend/handlers.py
```diff
@@ -0,0 +1,198 @@
...
+            k["position"], k["bounds"]
+        )
+        # Set indices that are greater than bounds
+        new_pos = k["position"]
+        new_pos[greater_than_bound[0]] = np.array(
+            [(ub[i] - lb[i]) * randr + lb[i] for randr,i in
+                (np.random.sample((k["position"].shape[0],)),
+                 k["position"].shape[0])]
+        )
+        new_pos[lower_than_bound[0]] = np.array(
```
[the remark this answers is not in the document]
whzup: Hmm, how do you think this'll work then? Just mask the values that are over the boundaries or mask the whole vector that is out-of-bounds? There might be some difference in behaviour.

pyswarms/backend/handlers.py
```diff
@@ -0,0 +1,198 @@
...
+        """
+        try:
+            lb, ub = k["bounds"]
+            bool_greater = k["position"] > ub
+            bool_lower = k["position"] < lb
+            new_pos = np.where(bool_greater, ub, k["position"]).where(
+                bool_lower, lb, k["position"]
+            )
+        except KeyError:
+            raise
```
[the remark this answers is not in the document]
whzup: Maybe something like this:
```
"Keyword 'position' or 'bounds' not found"
```

pyswarms/backend/handlers.py
```diff
@@ -0,0 +1,198 @@
...
+                    k["position"], k["bounds"], k["velocity"]
+            )
+
+            if not lower_than_bound and not greater_than_bound:
+                break
+
+            masking_vel = compute_velocity(k["swarm"], k["clamp"])
+            new_vel[lower_than_bound[0]] = masking_vel[lower_than_bound[0]]
+            new_vel[greate_than_bound[0]] = masking_vel[greater_than_bound[0]]
+        return new_vel
```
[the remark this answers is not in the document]
whzup: I see, how are we going to handle this though? Shall we use an `if` in the `operators` module to see if the strategy is "resample"?

whzup:
So the `self.memory` attribute stores the previous position. This means if the strategy is "shrink" or "intermediate" we do whatever with the positions and then store it in the `memory` attribute and use it in the next call. Seems like a good solution to me.

But I think this doesn't apply to the "resample" strategy. Maybe we should move the resample strategy to the `VelocityHandler` as it kind of breaks the API of the `BoundaryHandler`. The `VelocityHandler` will return a velocity anyways so it'd fit in there. The classification of "resample" as a position handling strategy doesn't make that much sense either to be honest. Shall I move it to the `VelocityHandler`?

Also, I think my current implementation is very computation heavy. If the first velocity is already out of bounds the next might be also out of bounds. This could lead to ridiculously huge computation counts just to recalculate one single velocity. We'd have to find a way to adjust the parameters like c1, c2 and w to have an efficient implementation. I'm sure there is a paper that talks about this.

I found another boundary handling strategy called "periodic" where we just use the `modulo` to reset the particle into the boundaries. It is probably quite simple and easy to implement. I'll implement it too so we have a good assortment of strategies :+1:

pyswarms/backend/handlers.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def __get_all_strategies(self):
+        """Helper method to automatically generate a dict of strategies"""
+        return {
+            k: v
+            for k, v in inspect.getmembers(self, predicate=inspect.isroutine)
+            if not k.startswith(("__", "_"))
+        }
+
+    def nearest(self, **k):
```
ljvmiranda921: Hi @whzup  

I realized that given a method:

```python
def nearest(pos, bounds):
    # do something
```

We can still pass a `**kwargs` argument such as `nearest(**kwargs)` as long as the dictionary is well-built. Can you check that one here? I think explicitly stating `posiiton` and `bounds` is better than just having an implicit `**kwargs`. There might be type errors in some special cases though but I think it would look better instead of indexing the kwargs dictionary for every parameter

whzup:
@ljvmiranda921 What do you think would be the best design for the integration into the `Optimizers`? We could pass it together with the bounds in a tuple like so:
```python
(np.array([0,0,0]), np.array([2,5,3]), "shrink")
```
Or separately in the `__init__` functions of the `Optimizers`. Maybe you can consider this in your Refactoring 🤔

whzup:
Ok, there is even more to it than I thought. We have to pass the strategy name to the `Optimizer` and then even to the `Topology` so they can use it for their `compute_pos` method. So I'd rather let you finish your refactoring of the `Optimizers` and `Topologies` so we don't get any merge conflicts, @ljvmiranda921 .

whzup:
Ok @ljvmiranda921 the things we have to decide on as soon as you're back are the follwing:
+ How does the user decide which position or velocity handling the PSO should use? Are you going to implement this interface to the user in the refactoring?
+ Shall we leave out the "resample" strategy? The current implementation seems a bit broken to me.
+ Shall we use explicit arguments for the methods like `shrink(pos, bounds)` or shall we leave it as it is?
+ It may be a good idea to move the `__out_of_bounds` method outside of the `BoundaryHandler` since the `VelocityHandler` also relies on the out-of-bounds-positions.
+ Is the error handling alright now?

ljvmiranda921:
> How does the user decide which position or velocity handling the PSO should use? Are you going to implement this interface to the user in the refactoring?

This won't happen in the refactor branch. It's better to solve this in another branch instead. Normally this should be passed in the `compute_position` or `compute_velocity`? Still fuzzy but I think this won't be a big problem.

> Shall we leave out the "resample" strategy? The current implementation seems a bit broken to me.

Yup, leave out the hard parts for now and let's just focus on the easier ones

> Shall we use explicit arguments for the methods like shrink(pos, bounds) or shall we leave it as it is?

I haven't seen your latest implementation yet but I think we can have explicit arguments? Either way they can be abstracted by passing a kwargs (i.e., `**{"pos": foo, "bounds": bar}`) right?

> It may be a good idea to move the __out_of_bounds method outside of the BoundaryHandler since the VelocityHandler also relies on the out-of-bounds-positions.

Sure, maybe add a mixins module inside the handlers directory?

> Is the error handling alright now?

There seems to be an error with Pyramid? I actually changed some tests in the refactor so we'll see.

whzup:
I found this error when working on the tests for the Handlers:
```python
>           self.rep.log.exception("Keyword 'position' or 'bounds' missing")
E           AttributeError: 'function' object has no attribute 'exception'
```
Seems weird to me :thinking: Do you know the cause of it @ljvmiranda921 ?

ljvmiranda921:
Where can I find that line, @whzup ?

whzup:
In the VelocityHandler strategies. It's the handling for a KeyError.

ljvmiranda921:
Check what `self.rep.log` is. It should be a logger, not a function. Shouldn't it be `self.rep.exception`?

whzup:
I think we forgot to add and `exception` method to the reporter :laughing:
```python
>           self.rep.exception("Keyword 'position' or 'bounds' missing")
E           AttributeError: 'Reporter' object has no attribute 'exception'
```

ljvmiranda921:
It can be solved by `self.rep.logger.exception`
https://docs.python.org/3/library/logging.html#logging.exception

whzup:
Hmm, there is a weird test failure in the kwargs test in the GeneralOptimizer. That is quite troublesome :frowning_face:

whzup:
Note: I'm going to make more or less trivial tests, only to see if the strategies do what they should do. This in combination with the test for exceeding particles should be enough for now.

ljvmiranda921:
Hey @whzup , this honestly seems to be a difficult problem hmmm...

whzup:
@ljvmiranda921, What do you think, shall I try to make some tests by calculating the desired positions by hand according to the algorithms or are the current tests sufficient?

ljvmiranda921:
Hi @whzup , I think the current tests for now are sufficient. If this is good then just ping me and we can merge this to `master`.

ljvmiranda921:
Merged! Thank you for your great contribution, @whzup !

PR #239, opened 2018-09-03
Analysis of Simple Circuit Added

pranav1698:
Added The Analysis of Simple Circuit Tutorial added 

## Description
I have made a bit of progress in the analysis of circuit tutorial please check and suggest changes!

### DIFF
--- a/examples/anaysis_of_simple_circuit.ipynb
+++ b/examples/anaysis_of_simple_circuit.ipynb
@@ -40,7 +40,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 17,
+   "execution_count": 16,
    "metadata": {},
    "outputs": [
     {
@@ -156,25 +156,21 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 30,
    "metadata": {},
    "outputs": [
     {
-     "ename": "NameError",
-     "evalue": "name 'pyswarms' is not defined",
-     "output_type": "error",
-     "traceback": [
-      "\u001b[0;31m---------------------------------------------------------------------------\u001b[0m",
-      "\u001b[0;31mNameError\u001b[0m                                 Traceback (most recent call last)",
-      "\u001b[0;32m<timed exec>\u001b[0m in \u001b[0;36m<module>\u001b[0;34m()\u001b[0m\n",
-      "\u001b[0;31mNameError\u001b[0m: name 'pyswarms' is not defined"
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "CPU times: user 7 µs, sys: 2 µs, total: 9 µs\n",
+      "Wall time: 14.3 µs\n"
      ]
     }
    ],
    "source": [
     "%%time\n",
     "\n",
-    "@pyswarms.cost\n",
     "def cost_func(I):\n",
     "    \"\"\"Method for definition of our cost function\n",
     "        Inputs:\n",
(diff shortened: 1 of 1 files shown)


ljvmiranda921:
You're doing great @pranav1698 ! So far the intro looks nice, we can edit this later after you've implemented the full notebook. Don't hesitate to message us if you have any questions :+1:

pranav1698:
The previous error was resolved,

whzup:
Hello @pranav1698, hope you're making progress! We noticed that you work on the master branch. That's nothing to worry about but the problem is that there exists no `@ps.cost` decorator in that branch (it's in the development branch and it's not the goal to have it included in the notebook). Switching the base branch of your feature branch is no big deal, though, since you haven't changed any existing files. Go through the following steps to do so:
1. Commit your work until now and copy your notebook file into another folder outside of pyswarms (just in case)

2. Update your development branch by using:
```
git fetch upstream
```
```
git checkout development
```
```
git merge upstream/development
```
3. Rebase your branch to use a different base branch:
```
git checkout analysis-of-simple-circuit
```
```
git rebase --onto development master analysis-of-simple-circuit
```

In case something goes wrong you can still use the backup from (1.) and make a new branch (with the base branch development) in which you can move it.
Note: The base branch of a newly created branch is the branch you have checked out when creating the new branch.

pranav1698:
@whzup I have tried the following but am getting the same error? 
I am not able to figure out what is causing this error...

whzup:
@pranav1698 So you have rebased your branch to have `development` as its base branch right? The problem is that you haven't vectorized your cost function. You just defined 2 functions which do not much on their own. Do the following:

1. Close this PR here.
2. Open a new PR where you compare the `development` branch of this repository to your branch.
3. Delete the code you copied from the cost decorator and use the built-in decorator (which should work if you are on the `development` branch)

This should resolve your error and the issue with the base branch. Hit me up after you've done this to see if all is fine then you can continue with the notebook 😄

pranav1698:
Just have a quick look, @whzup I think I am in the development branch here
screenshot from 2018-09-10 00-38-33

This is the snippet of my latest git log

whzup:
@pranav1698 Yes that's good! The problem is that this PR compares to the `master` branch of this project (excuse me I forgot to mention this earlier) and if we merge this PR we merge it into the `master` branch. We want to have it in the `development` branch, though. I don't know any method to change the branch a PR compares to so making a new PR is probably the easiest solution. So you can create a new PR where you choose the `development` branch of this repo on the left side of the comparision and your branch on the right side of the comparision.

ljvmiranda921:
Hi @pranav1698 ,

1. Make sure that you are passing a decorated version of your cost function in `optimizer.optimize()`:

Instead of passing this:

```python
def my_cost_func():
    return foo

optimizer.optimize(my_cost_func)
```

Pass this:

```python
@ps.cost
def my_cost_func()
    return foo

optimizer.optimize(my_cost_func)
```

2. Ensure that your cost function accepts a vector of size `d` (dimensions). The value of `d` is determined by your problem. How many dimensions is your problem trying to optimize? Make sure that your cost function can accept and compute that. 

I believe you're only optimizing for current I? (hence, `d=1`) Then that means your cost function should accept an array of size `1`, (e.g., `array([0.43])`) and output a float as the cost. It's important that your input is an array, because in this library, all particles, even if they only move in 1-dimension, are assumed to be vectors.

3. We recommend to use `numpy` library for your computations. It's much faster and optimized!

pranav1698:
Sorry @ljvmiranda921  I just followed the steps given by @whzup I didn't realised that you have changed the base branch into the develpoement branch!

ljvmiranda921:
Oops sorry, OK!

Issue #202, opened 2018-08-10
Make a Jupyter notebook tutorial for the analysis of a simple circuit


whzup:
## Goal
PSO can be utilized in a wide variety of fields. To broaden our collection of tutorials, we'd like to have another example where we analyse a simple circuit with PSO. For some inspiration, you can visit the [example section](https://github.com/ljvmiranda921/pyswarms/tree/master/examples). I propose that, for the beginning, we start by analysing the circuit shown below. It has a resistor and a diode. **The end-goal is to have a nice Jupyter notebook that goes through the whole process of writing this optimization program.**
## Method
![circuit](https://user-images.githubusercontent.com/39431903/43938822-29aaf9b8-9c66-11e8-8e54-01530db005c6.png)
As there are many models for diodes, let us use a more realistic one (a simplified Shockley equation) for this tutorial:

<a href="https://www.codecogs.com/eqnedit.php?latex=I&space;=&space;I_s&space;e^{\frac{v_D}{v_T}}" target="_blank"><img src="https://latex.codecogs.com/gif.latex?I&space;=&space;I_s&space;e^{\frac{v_D}{v_T}}" title="I = I_s e^{\frac{v_D}{v_T}}" /></a>

where:
 - <a href="https://www.codecogs.com/eqnedit.php?latex=I" target="_blank"><img src="https://latex.codecogs.com/gif.latex?I" title="I" /></a>:  diode current
- <a href="https://www.codecogs.com/eqnedit.php?latex=I_s" target="_blank"><img src="https://latex.codecogs.com/gif.latex?I_s" title="I_s" /></a>: reverse bias saturation current
- <a href="https://www.codecogs.com/eqnedit.php?latex=v_D" target="_blank"><img src="https://latex.codecogs.com/gif.latex?v_D" title="v_D" /></a>: diode voltage
- <a href="https://www.codecogs.com/eqnedit.php?latex=v_T" target="_blank"><img src="https://latex.codecogs.com/gif.latex?v_T" title="v_T" /></a>: thermal voltage (use 25.3 mV in this tutorial)

to use it in the tutorial I'd recommend to solve for <a href="https://www.codecogs.com/eqnedit.php?latex=v_D" target="_blank"><img src="https://latex.codecogs.com/gif.latex?v_D" title="v_D" /></a>:

[...]

PR #242, opened 2018-09-10
Added analysis of simple circuit

pranav1698:
## Description
Added analysis of single circuit

## Related Issue
Related to issue #202

### DIFF
--- a/examples/anaysis_of_simple_circuit.ipynb
+++ b/examples/anaysis_of_simple_circuit.ipynb
@@ -1,0 +1,293 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Analysing a Simple Circuit using Paricle Swarm Optimzation    ==============================\n",
+    "\n",
+    "In this example, we are going to use the `pyswarms` library to analyse a simple circuit by treating it as an optimization problem.The goal of the analysis of the circuit is to find out the current in the circuit.We will use the `pyswarms` library to find an *optimal* solution from a set of candidate solutions."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import sys\n",
+    "# Change directory to access the pyswarms module\n",
+    "sys.path.append('../')"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Running on Python version: 3.6.5 (default, Apr  1 2018, 05:46:30) \n",
+      "[GCC 7.3.0]\n"
+     ]
+    }
+   ],
+   "source": [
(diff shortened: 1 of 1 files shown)


examples/anaysis_of_simple_circuit.ipynb
```diff
@@ -0,0 +1,293 @@
...
+    "        I : current flowing in the diode\n",
+    "        \n",
+    "        Returns:\n",
+    "        ---------\n",
+    "        cost: our cost function\n",
+    "    \"\"\"\n",
+    "    cost = abs(9.24 - (0.069 * (math.log(I, 10))) - (100 * I))\n",
+    "    return cost\n",
+    "\n",
+    "def cost(cost_func):\n",
```
whzup: You can delete this function now and add the cost decorator above.

examples/anaysis_of_simple_circuit.ipynb
```diff
@@ -0,0 +1,293 @@
...
+     "text": [
+      "CPU times: user 10 µs, sys: 0 ns, total: 10 µs\n",
+      "Wall time: 13.8 µs\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%time\n",
+    "\n",
+    "def cost_func(I):\n",
```
whzup: Add the decorator before this function:
```python
@ps.cost
def cost_func(I):
    ...
```

examples/anaysis_of_simple_circuit.ipynb
pranav1698: I have to delete the "def cost(cost_func):"

whzup: The whole function you copied from the source files. This means this whole section:
>def cost(cost_func):
    """A decorator for the cost function
    This decorator allows the creation of much simpler cost functions. Instead of
    writing a cost function that returns a shape of :code:`(n_particles, 0)` it enables
    the usage of shorter and simpler cost functions that directly return the cost.
    A simple example might be:
     .. code-block:: python
         import pyswarms
         import numpy as np
         @pyswarms.cost
         def cost_func(x):
            cost = np.abs(np.sum(x))
            return cost
    The decorator expects your cost function to use a d-dimensional array (where
    d is the number of dimensions for the optimization) as and argument.
    .. note::
        Some :code:`numpy` functions return a :code:`np.ndarray` with single values in it.
        Be aware of the fact that without unpacking the value the optimizer will raise
        an exception.
    Parameters
    ----------
    cost_func : callable
        A callable object that can be used as cost function in the optimization
        (must return a :code:`float` or an :code:`int`).
    Returns
    -------
    cost_dec : callable
        The vectorized output for all particles as defined by :code:`cost_func`
    """
    def cost_dec(particles, **kwargs):
        n_particles = particles.shape[0]
        vector = np.array([cost_func(particles[i], **kwargs) for i in range(n_particles)])
        assert vector.shape == (n_particles, ), "The cost function should return a single value."
        return vector
    return cost_dec

We don't need it anymore if use the decorator from inside the package.

whzup:
Hey @pranav1698, hope you're doing fine!😄 How are you doing with the PR? We're not in a hurry I just wanted to check if everything's alright and see if you need any help 👍.

Issue #241, opened 2018-09-09
Refactor tests then add ABCTests


ljvmiranda921:
Sometimes we have repeating tests, i.e., tests that check the same condition:

```python
# test_module1.py

def test_boundary():
    # Some logic here
    assert foo == bar
```

In another module, we also have:

```python
# test_module2.py

def test_boundary():
    # Some logic here
    assert foo == bar
```

My idea is to create a base class, say, `ABCTestOptimizer`, and put all similar tests inside. Then, for each module, we simply inherit the base class then just add tests that are more specific whenever needed:

```python
# abc_optimizer.py

class ABCTestOptimizer(abc.ABC):
    def test_boundary():
         # Some logic here
         assert foo == bar
```

Then, 

```python
# test_module1.py

from abc_optimizer import ABCTestOptimizer

class TestModule1(ABCTestOptimizer):
    def test_something_specific_to_module1():
        # Some logic here
        assert foo == bar
```

We can maybe implement this technique in our high-level optimizers and topologies

Issue #240, opened 2018-09-09
Refactor error handling from LBYL to EAFP


ljvmiranda921:
Right now, most of our optimizers use the `assertions()` method to catch invalid inputs. This method is implemented in a Look Before You Leap (LBYL) fashion. However, idiomatic Python is often written in an Easy to Ask Forgiveness than Permission (EAFP) design, so it's better to catch errors using that.

This also removes the need to be very strict with the types passed in `bounds` or `velocity_clamp`. We usually require a `tuple` of `lists`, but in reality, an `iterable` should suffice, please check those conditions when dealing with this issue.

My idea is to delegate error handling in the backend, rather than having an `assertions()` method for each optimizer. Perhaps, it might even be better to have an `assertions`module and do all checks there. Hopefully, this would make our implementation much cleaner.

whzup:
The `Searches` also use this type of asserting. Maybe you want to refactor them as well.

PR #243, opened 2018-09-10
SPRINT: Refactor parts of the codebase

ljvmiranda921:
## Description
A large pull request that deals mainly with refactoring some parts of the codebase. Specifically:
- The tests for topology and high-level optimizers
- Error-handling from LBYL to EAFP
- Remove if-else hierarchy in General Optimizer

## Related Issues
#232, #240, #241

## Motivation and Context
It is important that we make our code scalable, and has a consistent design before we add bigger features.

## Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/pyswarms/backend/topology/random.py
+++ b/pyswarms/backend/topology/random.py
@@ -18,18 +18,23 @@ from .base import Topology
 
 
 class Random(Topology):
-    def __init__(self, static=False):
+    def __init__(self, k, static=False):
         """Initializes the class
 
         Parameters
         ----------
+        k : int
+            number of neighbors to be considered. Must be a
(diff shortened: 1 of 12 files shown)


ljvmiranda921:
## Refactor GeneralOptimizer

The idea is to change `compute_gbest(self, swarm)` into something like `compute_gbest(self, swarm, **kwargs)` so that all other params such as `p`, `k`, and `r` can be relegated locally to the topologies.
From this, you can just use one API call when calling `compute_gbest` here: https://github.com/ljvmiranda921/pyswarms/blob/development/pyswarms/single/general_optimizer.py#L260

PR #244, opened 2018-09-12
SPRINT: Update general optimizer

ljvmiranda921:
Change topology API so that GeneralOptimizer will only have one consistent API call


ljvmiranda921:
One thing I'm planning to do is have all options such as `p`, `k`, and `r` be required during initialization itself, rather than being called in `compute_gbest()`. These values do not change (except the number of neighbors in `VonNeumann`, but that can easily be handled), so there's no problem setting them as attributes.

PR #247, opened 2018-09-12
Refactor: Abstract Tests

ljvmiranda921:
## Goal

Error-checking of passed arguments should be delegated to the `backend`. If it's going to be caught by the backend, let it be caught by the backend. No need to catch them on the optimizer side because it becomes redundant.


ljvmiranda921:
These kind of tests should be caught by the backend itself (and other dependency modules)

```
@pytest.mark.parametrize(
    "options",
    [{"c2": 0.7, "w": 0.5}, {"c1": 0.5, "w": 0.5}, {"c1": 0.5, "c2": 0.7}],
)
def test_keyword_exception(options):
    """Tests if exceptions are thrown when keywords are missing"""
    with pytest.raises(KeyError):
        GlobalBestPSO(5, 2, options)


@pytest.mark.parametrize(
    "bounds",
    [
        tuple(np.array([-5, -5])),
        (np.array([-5, -5, -5]), np.array([5, 5])),
        (np.array([-5, -5, -5]), np.array([5, 5, 5])),
    ],
)
def test_bounds_size_exception(bounds, options):
    """Tests if exceptions are raised when bound sizes are wrong"""
    with pytest.raises(IndexError):
        GlobalBestPSO(5, 2, options=options, bounds=bounds)


@pytest.mark.parametrize(
    "bounds",
    [
        (np.array([5, 5]), np.array([-5, -5])),
        (np.array([5, -5]), np.array([-5, 5])),
    ],
)
def test_bounds_maxmin_exception(bounds, options):
    """Tests if the max bounds is less than min bounds and vice-versa"""
    with pytest.raises(ValueError):
        GlobalBestPSO(5, 2, options=options, bounds=bounds)


@pytest.mark.parametrize(
    "bounds",
    [
        [np.array([-5, -5]), np.array([5, 5])],
        np.array([np.array([-5, -5]), np.array([5, 5])]),
    ],
)
def test_bound_type_exception(bounds, options):
    """Tests if exception is raised when bound type is not a tuple"""
    with pytest.raises(TypeError):
        GlobalBestPSO(5, 2, options=options, bounds=bounds)


@pytest.mark.parametrize("velocity_clamp", [(1, 1, 1), (2, 3, 1)])
def test_vclamp_shape_exception(velocity_clamp, options):
    """Tests if exception is raised when velocity_clamp's size is not equal
    to 2"""
    with pytest.raises(IndexError):
        GlobalBestPSO(5, 2, velocity_clamp=velocity_clamp, options=options)


@pytest.mark.parametrize("velocity_clamp", [(3, 2), (10, 8)])
def test_vclamp_maxmin_exception(velocity_clamp, options):
    """Tests if the max velocity_clamp is less than min velocity_clamp and
    vice-versa"""
    with pytest.raises(ValueError):
        GlobalBestPSO(5, 2, velocity_clamp=velocity_clamp, options=options)


@pytest.mark.parametrize("err, center", [(IndexError, [1.5, 3.2, 2.5])])
def test_center_exception(err, center, options):
    """Tests if exception is thrown when center is not a list or of different shape"""
    with pytest.raises(err):
        GlobalBestPSO(5, 2, center=center, options=options)

```

ljvmiranda921:
## Task list

- [ ] Remove `assertions` in high-level optimizers
- [ ] Remove `if-else` error catching in `GeneralOptimizer`
- [ ] Add more assertion checks in Topologies and/or in the backend

ljvmiranda921:
There seems to be a problem in Python 3.4 regarding $DISPLAY in the plotters test. I did not change the test themselves but ran isort on the file. Maybe the ordering of imports mattered? Sometimes I just want to drop py34 altogether, so many API changes between 3.4 and 3.5...

ljvmiranda921:
Things to do:
- Solve py34 problem
- Add Raises in some docstrings

ljvmiranda921:
You might be interested in my progress @whzup , just check if you have timep

pyswarms/single/general_optimizer.py
```diff
@@ -88,36 +90,36 @@ def __init__(
             number of particles in the swarm.
         dimensions : int
             number of dimensions in the space.
-        options : dict with keys :code:`{'c1', 'c2', 'w'}` or :code:`{'c1', 'c2', 'w', 'k', 'p'}`
+        options : dict with keys :code:`{'c1', 'c2', 'w'}` or :code:`{'c1',
+                'c2', 'w', 'k', 'p'}`
             a dictionary containing the parameters for the specific
             optimization technique.
-                * c1 : float
+                * c5 : float
```
whzup: It should say c1 here, shouldn't it?

ljvmiranda921: Yes you're correct. Can you also help me find why the test is currently failing? Seem to be running out of ideas...

ljvmiranda921: Also, thanks for reviewing this work. Please check the docstrings and other stuff so that I can edit it as we go. Thanks a lot!

ljvmiranda921:
OK, so I think I have solved the problem..

It turns out that `conftest.py` also imports some modules for `plotters`, and that implicitly imports `pyplot` first before setting `Agg` at runtime, causing the `tkinter` errors. This may have caused by running `isort` into `conftest.py`. I was only editing `test_plotters` without realizing that `conftest` is also affected.

I have solved this by rearranging the imports in `conftest`, and adding an `isort` skip in the file's docstring

ljvmiranda921:
If this is good, I can rebase and merge this to general, then maybe we can proceed to the next set of tasks. :+1:

whzup:
Nice! Glad you found the error :smile: I'll try to have a look this evening :+1:

whzup:
@ljvmiranda921 Looks good to me! I like that fact that the PR removes more code than it adds :smiley:

PR #249, opened 2018-09-14
fix bug: add parameter init_pos

dfhljf:


Kosisochi:
Hello, how do I specify if I want to either maximize or minimize my objective function?

Issue #250, opened 2018-09-17
Ring topology update velocity basing on gbest instead of lbest.


Kevin9436:
**Describe the bug**
In backend.topology.ring module, the compute_gbest() method has already calculate the best_neighbor, but in the compute_velocity() method it call operators.compute_velocity(), which uses gbest for updating.

**Expected behavior**
Pass the best_neighbor to the ring.comput_velocity() method and calculate the social part with it. 

**Additional context**
The lbest algorithm is mentioned by you in issue[#170](https://github.com/ljvmiranda921/pyswarms/issues/170)
Reference: A. Engelbrecht, Computational Intelligence 2nd Ed. p290,292,293

@whzup

whzup:
Welcome, @Kevin9436! Thank you for reporting this bug!

Excuse me, though, as I don't fully understand your point here 😞. So you say we should pass the `best_neighbor` to the `compute_velocity()` method and calculate the social part with it. We actually already do that but with a bit of a detour to the `LocalBestPSO`/`GeneralOptimizerPSO`:
We store the `best_neighbor` in the swarm (using the `swarm.best_pos` attribute) in [this line](https://github.com/ljvmiranda921/pyswarms/blob/master/pyswarms/single/local_best.py#L218) and then use it in the `compute_velocity()` method [here](https://github.com/ljvmiranda921/pyswarms/blob/master/pyswarms/backend/operators.py#L123). If this isn't what you reported please correct me!

Kevin9436:
My confusion is about [this line](https://github.com/ljvmiranda921/pyswarms/blob/b83e99be17d3b05a5948b73ad11fd2a0f6b0ce36/pyswarms/backend/topology/ring.py#L83). If my prehension is not wrong, this `best_pos` is the global best of the whole swarm and this line is to pick the minimum cost position from all best neighbour. 

Since the neighbour size of each particle is `k`, for local best algorithm, the particle can only know the best position of the nearest `k` particles, while you give the global best position to it.

I think the right way to do this is to return a new parameter `local_best = swarm.pbest_pos[best_neighbor]` and replace the `best_pos` [here](https://github.com/ljvmiranda921/pyswarms/blob/b83e99be17d3b05a5948b73ad11fd2a0f6b0ce36/pyswarms/backend/operators.py#L123) with it.

whzup:
[...]

PR #253, opened 2018-09-19
Fix bug in topologies

whzup:
## Description

I changed the way the topologies compute the `best_pos` value.

## Related Issue

Resolves: #250 

## Motivation and Context

As pointed out in #250 the topologies do not calculate the best position of their neighbours but the global best position which goes against the purpose of the topologies.

## How Has This Been Tested?

Tests

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:


whzup:
There is a problem here as it seems that the Ring topology doesn't converge fast enough for the `test_objective_func_with_kwargs` tests. So we have failing tests because the final cost is too big. Maybe this can be fixed by smartly picking the options :thinking:

ljvmiranda921:
> There is a problem here as it seems that the Ring topology doesn't converge fast enough (which actually is what is to be expected) for the test_objective_func_with_kwargs tests. So we have failing tests because the final cost is too big. Maybe this can be fixed by smartly picking the options thinking

Hi @whzup , yup that can be arranged. I'm sorry but I thought the change in my refactor branch will be small and we might have a lot of merge conflicts when this is merged to master. Wondering if you can **PR this on my refactor branch instead?**  (`refactor/general`)

For now, just note the changes in the topologies then we can do a safe merge afterwards :+1:

pyswarms/single/local_best.py
```diff
@@ -251,12 +251,8 @@ def optimize(self, objective_func, iters, fast=False, **kwargs):
             )
         # Obtain the final best_cost and the final best_position
         final_best_cost = self.swarm.best_cost.copy()
-        final_best_pos = self.swarm.best_pos.copy()
-        # Write report in log and return final cost and position
-        self.rep.log(
-            "Optimization finished | best cost: {}, best pos: {}".format(
-                final_best_cost, final_best_pos
-            ),
-            lvl=20,
```
whzup: @ljvmiranda921 which logging is right here? The one that I removed or the one that I added? It was a merge conflict and I just used the one from my commits that may be wrong :confused:

whzup:
@ljvmiranda921 what shall we do about the failing test in the `test_objective_func_with_kwargs`? The tweaking of the options does not seem to work :confused:

ljvmiranda921:
Let’s leave that for now, maybe we can relax the assert later on? We’re not really testing for success, just that the values pass to the objective function as expected.

whzup:
I see. #238 might even solve this issue to be honest 😄

whzup:
@ljvmiranda921 Do you have a better idea than hardcoding the results into the tests? It's very tedious and there are a lot of special cases that have to be covered :confused:

whzup:
**Note:** I tried to check whether or not the positions that the topologies return are correct by examining the `neighbor_idx` attribute. It turned out that it is not accessible. When I tried to print the attribute inside of the test in the `test_pyramid.py` file this was the error that it returned:
```python
>       print(topology.neighbor_idx)
E       AttributeError: type object 'Pyramid' has no attribute 'neighbor_idx'
```
Maybe we need to use an abstract attribute for the `neighbor_idx`.  See [here](https://stackoverflow.com/questions/23831510/abstract-attribute-not-property?noredirect=1&lq=1) for a guide.

whzup:
@ljvmiranda921 This one works now more or less. The tests now look if in the best position matrix (where the positions of the best neighbours of every particle is stored) either contains the best position or the second best position that exists. I think testing the whole matrix is just a waste of space and time to be honest... What do you think?
The only thing that still inflicts errors is this line in the `test_general_optimizer.py` file:
```python
topologies = [topo for _, topo in inspect.getmembers(t, istopology)]
```
It doesn't seem to return the class instances :confused:. I'm not quite sure how to interpret the [docs](https://docs.python.org/3.6/library/inspect.html#inspect.getmembers) either.

ljvmiranda921:
Hi @whzup ! Sorry for being MIA for a long time, work got in and I had no time working on pyswarms. I'll check this weekend or the next :+1:

whzup:
@ljvmiranda921 No problem I had a lot to do as well 😄. I found the culprit it seems that some parenthesis vanished. It should say:
```python
topologies = [topo() for _, topo in inspect.getmembers(t, istopology)]
```
Now two failing tests are remaining until this PR is finished 😆.

whzup:
Well got it at last... Had to skip some tests tho :confused:

ljvmiranda921:
Hey @whzup ! 

Sorry for the very late thing. Anyway, this looks good already. I'm gonna approve this and just wait for you to to rebase everything so that I can merge to the refactor branch!

ljvmiranda921:
@all-contributors please add @whzup for maintenance, code, doc, tests, ideas, and review

PR #263, opened 2019-01-28
Update how dependencies are created

ljvmiranda921:
Now, we're using `pip-tools` as a way to manage all our dependencies. We need to set it up first using an `.in` file, then compile it via `make requirements.txt` or `make requirements-dev.txt`. In this manner, all dependencies for the project are pinned explicitly.

Reference: #262

## Types of changes

## Checklist:


ljvmiranda921:
`matplotlib==3.0.2`, the version we actually need, cannot be installed in py34. Let's remove support for that!

Issue #258, opened 2018-11-22
Parallel particle evaluation


soamaven:
**Is your feature request related to a problem? Please describe.**
I know I have an expensive function to evaluate, I am doing a simulation based optimization of a hyper parameter space. I am planning to use HPC environment because of this. Can particles be evaluated in parallel? Synchronously or even asynchronously? I thought an [approach similar to this](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC1769316/) would be interesting (let me know if article is inaccessible).

**Describe the solution you'd like**
A switch to enable threads for particle evaluations. In particular, I am fine with firing off `os.system()` commands and waiting for return. 

**Describe alternatives you've considered**
I supposed I could write a sub-script that executes the swarm in parallel, evaluates the target function for each particle, and returns those values to the optimizer. This wouldn't allow for async optimization. Also, I could also try to write another optimizer, one that takes an option for the number of parallel particles.

What is the best way to go here? Still digging into this API currently.

ljvmiranda921:
Hi @soamaven , 

Thank you for this issue. To be honest I've been meaning to add this feature [since last August](https://github.com/ljvmiranda921/pyswarms/wiki/Development-Roadmap), especially that I see a lot of use-cases involving very expensive objective functions. 

I would love to work on this, it's just that we're still finding time (especially we have a backlog of refactor issues to do) to dev (most of us have day jobs or at uni). Hopefully I can work on this on Q1 or Q2 next year, we just need to make a new release that solves the past issues.

If you can help us develop a module or base framework for this, then we'd really appreciate it! :smile:

soamaven:
Thanks!

Thinking out loud a bit below, let me know where I am misunderstanding the code.:

So it looks like it would be easiest to just implement a custom optimization loop! So its probably easiest to write an optimization loop that parallelizes calculating the `numpy.ndarray` cost in parallel, either in an async or sync fashion, where in the async loop, it would be a nested iteration/async loop. 

I am sure there is demand for making parallelization transparent though. My thoughts on that are probably things you are aware of.

It looks like an inheritable and simple `multiprocessing.pool.map()` implementation within the optimizer base might be hard to do since the general `SwarmOptimizer` requires each child to implement `optimize`. It also actually looks like there is some code repetition between the different optimizers that could maybe be done away with by implementing those common portions of the algorithm (update positions, update bests, etc.) of the swarm loop higher in the hierarchy.

[...]

PR #312, opened 2019-02-26
Support parallel particle evaluation

danielcorreia96:
## Description

The goal of this PR is to support parallel particle evaluation in PySwarms.

~~### **This is a draft PR to collect feedback on implementation specific decision regarding this feature.**~~

~~At the moment, I only updated the optimize() implementation to use the `compute_objective_function` abstraction, but the same changes can be applied easily to the remaining optimizers.~~

## Related Issue

#258 

## Motivation and Context

The main motivation for this PR is to improve the performance of PySwarms optimizers by parallelizing the evaluation of each particle during optimization.

This is especially relevant when: 
1. The objective function is expensive to compute.
2. The swarm has a high number of particles (thousands, millions, ...)

My proposal at the moment consists in abstracting the existing direct calls on the `objective_func` to a new method `compute_objective_function` that controls whether the evaluation should be done using parallelization or not.

Keep in mind that, since Python has a GIL we can't parallelize using threads and are forced to use `multiprocessing` to parallelize the computation across multiple processes.

Moving on the actual changes, there are some highlights that I'd like to explain:
1. Constructing a `multiprocessing.Pool` before starting the optimization iterations
   - Starting up new processes in every iteration for parallelization is expensive
   - We can just reuse the processes controlled by the Pool in every iteration
2. Passing `n_processes` in optimize method vs optimizer constructor
   - We may want to run the same optimizer configuration with varying degrees of parallelization.
   - This also further supports the idea of creating the Pool in the `optimize` method instead of during construction of the optimizer instance.
3. `partial(objective_func, **kwargs)` magic:
   - This is needed to support objective functions with arguments.

### **There is one small drawback at the moment** ⚠️ 
The objective functions built using the `cost` decorator are not supported in parallel execution scenarios. 

This happens because the `cost_dec` function is local to the decorator module and the module that starts up the Pool doesn't know about it, which makes the pickling process fail.

## How Has This Been Tested?

At the moment, only a personal project that uses binary PSO and an expensive objective function.

## Types of changes

## Checklist:

### DIFF
--- a/pyswarms/backend/operators.py
+++ b/pyswarms/backend/operators.py
@@ -16,6 +16,7 @@ import numpy as np
 
 from ..utils.reporter import Reporter
 from .handlers import BoundaryHandler, VelocityHandler
+from functools import partial
 
 
 rep = Reporter(logger=logging.getLogger(__name__))
@@ -207,3 +208,32 @@ def compute_position(swarm, bounds, bh):
         raise
     else:
         return position
+
+
+def compute_objective_function(swarm, objective_func, pool=None, **kwargs):
+    """Evaluate particles using the objective function
+
+    This method evaluates each particle in the swarm according to the objective function passed.
+
+    If a pool is passed, then the evaluation of the particles is done in parallel using multiple processes.
+
+    Parameters
+    ----------
+    swarm : pyswarms.backend.swarms.Swarm
+        a Swarm instance
+    objective_func : function
+        objective function to be evaluated
+    pool: multiprocessing.Pool
+        multiprocessing.Pool to be used for parallel particle evaluation
+    kwargs : dict
+        arguments for the objective function
+    Returns
+    -------
+    numpy.ndarray
+        Cost-matrix for the given swarm
+    """
+    if pool is None:
+        return objective_func(swarm.position, **kwargs)
+    else:
+        results = pool.map(partial(objective_func, **kwargs), np.array_split(swarm.position, pool._processes))
+        return np.concatenate(results)
--- a/tests/optimizers/abc_test_optimizer.py
+++ b/tests/optimizers/abc_test_optimizer.py
@@ -83,6 +83,13 @@ class ABCTestOptimizer(abc.ABC):
         opt.optimize(sphere, 2000)
         assert np.array(opt.cost_history).shape != (2000,)
 
+    def test_parallel_evaluation(self, obj_without_args, optimizer, options):
+        """Test if parallelization breaks the optimization process"""
+        import multiprocessing
+        opt = optimizer(100, 2, options=options)
+        opt.optimize(obj_without_args, 2000, n_processes=multiprocessing.cpu_count())
+        assert np.array(opt.cost_history).shape == (2000,)
+
     def test_obj_with_kwargs(self, obj_with_args, optimizer, options):
         """Test if kwargs are passed properly in objfunc"""
         x_max = 10 * np.ones(2)
--- a/pyswarms/base/base_discrete.py
+++ b/pyswarms/base/base_discrete.py
@@ -129,7 +129,7 @@ class DiscreteSwarmOptimizer(abc.ABC):
         self.velocity_history.append(hist.velocity)
 
     @abc.abstractmethod
-    def optimize(self, objective_func, iters, **kwargs):
+    def optimize(self, objective_func, iters, n_processes=None, **kwargs):
         """Optimize the swarm for a number of iterations
 
         Performs the optimization to evaluate the objective
@@ -142,6 +142,8 @@ class DiscreteSwarmOptimizer(abc.ABC):
             objective function to be evaluated
         iters : int
             number of iterations
+        n_processes : int
+            number of processes to use for parallel particle evaluation (default: None = no parallelization)
         kwargs : dict
             arguments for objective function
 
--- a/pyswarms/base/base_single.py
+++ b/pyswarms/base/base_single.py
@@ -129,7 +129,7 @@ class SwarmOptimizer(abc.ABC):
         self.velocity_history.append(hist.velocity)
 
     @abc.abstractmethod
-    def optimize(self, objective_func, iters, **kwargs):
+    def optimize(self, objective_func, iters, n_processes=None, **kwargs):
         """Optimize the swarm for a number of iterations
 
         Performs the optimization to evaluate the objective
@@ -142,6 +142,8 @@ class SwarmOptimizer(abc.ABC):
             objective function to be evaluated
         iters : int
             number of iterations
+        n_processes : int
+            number of processes to use for parallel particle evaluation (default: None = no parallelization)
         kwargs : dict
             arguments for objective function
 
--- a/pyswarms/discrete/binary.py
+++ b/pyswarms/discrete/binary.py
@@ -56,8 +56,9 @@ import logging
 
 # Import modules
 import numpy as np
+import multiprocessing as mp
 
-from ..backend.operators import compute_pbest
+from ..backend.operators import compute_pbest, compute_objective_function
 from ..backend.topology import Ring
 from ..backend.handlers import BoundaryHandler, VelocityHandler
 from ..base import DiscreteSwarmOptimizer
@@ -134,7 +135,7 @@ class BinaryPSO(DiscreteSwarmOptimizer):
         self.vh = VelocityHandler(strategy=vh_strategy)
         self.name = __name__
 
-    def optimize(self, objective_func, iters, **kwargs):
+    def optimize(self, objective_func, iters, n_processes=None, **kwargs):
         """Optimize the swarm for a number of iterations
 
         Performs the optimization to evaluate the objective
@@ -146,6 +147,8 @@ class BinaryPSO(DiscreteSwarmOptimizer):
             objective function to be evaluated
         iters : int
             number of iterations
+        n_processes : int
+            number of processes to use for parallel particle evaluation (default: None = no parallelization)
         kwargs : dict
             arguments for objective function
 
@@ -163,11 +166,14 @@ class BinaryPSO(DiscreteSwarmOptimizer):
         # Populate memory of the handlers
         self.vh.memory = self.swarm.position
 
+        # Setup Pool of processes for parallel evaluation
+        pool = None if n_processes is None else mp.Pool(n_processes)
+
         self.swarm.pbest_cost = np.full(self.swarm_size[0], np.inf)
         for i in self.rep.pbar(iters, self.name):
             # Compute cost for current position and personal best
-            self.swarm.current_cost = objective_func(
-                self.swarm.position, **kwargs
+            self.swarm.current_cost = compute_objective_function(
+                self.swarm, objective_func, pool, **kwargs
             )
             self.swarm.pbest_pos, self.swarm.pbest_cost = compute_pbest(
                 self.swarm
--- a/pyswarms/single/general_optimizer.py
+++ b/pyswarms/single/general_optimizer.py
@@ -61,8 +61,9 @@ import logging
 
 # Import modules
 import numpy as np
+import multiprocessing as mp
 
-from ..backend.operators import compute_pbest
+from ..backend.operators import compute_pbest, compute_objective_function
 from ..backend.topology import Topology
 from ..backend.handlers import BoundaryHandler, VelocityHandler
 from ..base import SwarmOptimizer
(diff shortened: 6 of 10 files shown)


whzup:
@danielcorreia96 Just leave a comment here as soon as you have finished working on this and I'll review it😄

danielcorreia96:
@whzup You can already review the current set of changes. 

The rest of the work is just propagating these changes to the other optimizers and updating docstrings.

I would just like make sure the main idea is good enough before expanding the changes to rest of the optimizers 😄

danielcorreia96:
Ok, I'll propagate the changes to the remaining optimizers and maybe add a test case (using the fixture decorator) for this feature 😄 

Regarding this comment:

>  I think we have to state in the documentation that one has either to use the `@cost` decorator to create a cost function or use the `X.shape[0]`. If they just use the number of particles to vectorize their cost function this might go wrong with multiprocessing.
> 

I'm not sure we're on the same page here. According to my attempts, using multiprocessing + `@cost` decorator doesn't work. You can find some notes on this problem in these links: 
- http://gael-varoquaux.info/programming/decoration-in-python-done-right-decorating-and-pickling.html
- https://stackoverflow.com/questions/26576550/using-multiprocessing-with-a-decorated-function-results-in-a-picklingerror

Can you clarify what you meant by this (and the `X.shape[0]`part)? 🙏

danielcorreia96:
cc: @whzup 
I've finished propagating the changes to the optimizers. 
I also added a test case to check if parallelization breaks optimization.

I'm not sure if the documentation I added was sufficient or not though 😆

whzup:
@danielcorreia96 Excuse me I completely forgot about this PR 😄

> I'm not sure we're on the same page here. According to my attempts, using multiprocessing + `@cost` decorator doesn't work.

I meant that maybe someone would have the idea to just vectorize his objective function using the number of particles instead of using `X.shape[0]`. Thus this code:
```python
def cost(X):
    #stuff 
    n_part = 200
    cost = [calc_cost(X[i]) for i in range(n_part)]
    return cost
```

wouldn't work since the multiprocessing splits up the `X`, right? It may not occur often but it safer to have it in the docs anyway.

I see your problem with the decorator. Since we already make changes related to the multiprocessing we can also include a change on the `cost` decorator if you're up for it.

danielcorreia96:
> I meant that maybe someone would have the idea to just vectorize his objective function using the number of particles instead of using X.shape[0]. Thus this code:
> ...
> wouldn't work since the multiprocessing splits up the X, right? It may not occur often but it safer to have it in the docs anyway

Yes you're right that multiprocessing wouldn't work in this scenario. Although one could argue that the code was already wrong from the start, since we were using indices/range to iterate on the particles 🤣

pyswarms/backend/operators.py
```diff
@@ -207,3 +208,35 @@ def compute_position(swarm, bounds, bh):
...
+
+
+def compute_objective_function(swarm, objective_func, pool=None, **kwargs):
+    """Evaluate particles using the objective function
+
+    This method evaluates each particle in the swarm according to the objective function passed.
+
+    If a pool is passed, then the evaluation of the particles is done in parallel using multiple processes.
+
+    Note: due to pickling limitations, if the pyswarms.cost decorator is used to build the objective function,
```
danielcorreia96: @whzup I have this docstring here to highlight the incompability of multiprocessing and the `cost` decorator. Should I also add this docstring to `cost` decorator module?

danielcorreia96:
cc: @whzup @ljvmiranda921 
I've opened this PR for review (from draft/WIP state) since it looks to be stable enough.

I'll try to update the branch and clean up the commit messages this weekend 👍

pyswarms/backend/operators.py
whzup: Would it be a lot of work to just fix the decorator such that it works? Seems there are some solutions in the links you provided above.

danielcorreia96: I tried using those solutions before but I couldn't make it work...

Today, I tried again and it worked 🤣 I think I was forgetting to put the `cost_func` argument inside the `@wraps` decorator from functools.

I've updated the PR to contain this fix 👍

ljvmiranda921:
Hi, @danielcorreia96 and @whzup 
Let me review this next weekend. Sorry, work came in!

tests/optimizers/abc_test_optimizer.py
```diff
@@ -81,7 +81,14 @@ def test_ftol_effect(self, options, optimizer):
         """Test if setting the ftol breaks the optimization process"""
         opt = optimizer(10, 2, options=options, ftol=1e-1)
         opt.optimize(sphere, 2000)
-        assert np.array(opt.cost_history).shape != (2000,)
+        assert np.array(optimizer.cost_history).shape != (2000,)
```
whzup: Just out of curiosity, why did you change this line?🤔 Shouldn't it stay as it is?😄

ljvmiranda921:
Sorry for the delays, @danielcorreia96 and @whzup .
Give me a hard deadline of end of March. If I still can't do the review, then @whzup can merge this PR.

tests/optimizers/abc_test_optimizer.py
danielcorreia96: Hm you're right. I probably didn't notice it because this test fixture is skipped :smile: 

I'll update the PR with the original code.

danielcorreia96: @whzup I think it's ok now 👍

whzup:
@ljvmiranda921 Is it correct that we want merge this into the `master` branch? Just to make sure that we don't accidently mess up something 😄

ljvmiranda921:
@whzup Yes merge to master! Cleanup the commits first and then do a “Rebase and merge” or “Squash and merge”

ljvmiranda921:
Hey @whzup , is this good to go? I can merge this to master on the weekend cc: @danielcorreia96

whzup:
@ljvmiranda921 Yes this should be alright :+1:

ljvmiranda921:
Merged! Thank you for your awesome contribution, @danielcorreia96 , really sorry merging this took so long. Will make a minor release next week with this feature! Again, thank you so much!

msat59:
Hi.

Parallel processing only does not work in Windows machine

ljvmiranda921:
Hi @msat59 , as of now pyswarms is not tested (and not guaranteed to work) on the Windows operating system. But I'll still try to find a way on how to solve that

pznaseri:
I imported pyswarms and would like to use the "single.GlobalBestPSO" module. This is my loss function that I'd like to compute in parallel for particles of each iteration. I tried setting up "n_processes" as 20. I can see that at the beginning 20 CPU start working but it stops immediately and the optimization just get stuck. 
Can you please help? 
```
def lossfunction_GSM(x, desired_sparam):
    predMag = np.zeros((nfreq,1),dtype=float)
    S12TETE_sim = np.zeros((nfreq,1),dtype=float)
    dist = np.mean(abs(x -  latentspace[0]), axis = 1)
    minDistance = dist[np.argmin(dist)]
    if minDistance < 0.02: #use the NN models to evaluate the particle's [S] 
        for j in range(nfreq):
            freq = float(0 + j * float((1-0) / (nfreq-1)))
            predMag[j] = MAGpredictorTE.predict([np.array([freq]), np.transpose(np.expand_dims(x,axis=-1))])   
            print('Using NN model!')
            loss = np.mean(np.square(abs(predMag- desired_sparam))[:,0])
            return loss 
    #looking if the feature is close to some input feature so we only need to call cascade
    else:
        cascadeThickness = 1.575 #denormalizing the thickness 
        particleGSMs= decoderEMMS.predict(np.transpose(np.expand_dims(x,axis=-1)))
        GSM1, GSM2 = np.split(particleGSMs[0],[original_dim])
        GSM1re = GSM1.reshape(digit_size, int(digit_size/2))
        GSM2re = GSM2.reshape(digit_size, int(digit_size/2))
        distImage1 = np.mean(np.mean(abs(GSM1re.reshape(52,26) - images), axis =-1),axis =-1)
        distImage2 = np.mean(np.mean(abs(GSM2re.reshape(52,26) - images), axis =-1),axis =-1)
        
        prticleGeometry1 = singleLayerImages[np.argmin(distImage1)]
        GSMfile1 = '%s//%s' % (singlelayerGSMs,prticleGeometry1.replace('CS-0-','CS-').replace('.png','.h5'))
        prticleGeometry2 = singleLayerImages[np.argmin(distImage2)]
        GSMfile2 = '%s//%s' % (singlelayerGSMs,prticleGeometry2.replace('CS-0-','CS-').replace('.png','.h5'))
        print('GSMfile1:{}'.format(GSMfile1))
        print('GSMfile2:{}'.format(GSMfile2))
        gem1 = prticleGeometry1.replace('CS-0-','CS-').replace('.png','')
        gem2 = prticleGeometry2.replace('CS-0-','CS-').replace('.png','')
        fileName =duallayerAddedPath+'//'+gem1 +'-1.575-' +gem2+'.csv'
        if os.path.exists(fileName):
            print('reading from file')
            data = pd.read_csv(resultfile) 
            sparams = data.values
            for n in range(sparams.shape[1]):
                S12TETE_sim[n] = abs(complex(sparams[0,n]))
        else:  
            print('calling cascade!')
            S12TETE_sim = call_cascade(GSMfile1, cascadeThickness*1e-3,GSMfile2)
        loss = np.mean(np.square(abs(abs(S12TETE_sim)- desired_sparam))[:,0]) 
        return loss
 ```
```
def f(x):
    n_particles = x.shape[0]
    #print(n_particles)
    j = [lossfunction_GSM(x[i], desired_sparam) for i in range(n_particles)]
    return np.array(j)
```
```
dimensions = latent_dim 
bounds = (np.squeeze(minZmean), np.squeeze(maxZmean))

# Initialize swarm
options = {'c1': 1.5, 'c2': 1.3, 'w':0.5}

# Call instance of PSO

optimizer = pyswarms.single.GlobalBestPSO(n_particles=30, dimensions=dimensions, options=options, bounds=bounds)

# Perform optimization
cost, pos = optimizer.optimize(f, iters=100)
plot_cost_history(cost_history=optimizer.cost_history)
plt.show()
```

**Comment by lulelita** [comment]
Hi there,
 I am running a pretty complex routine and the last bit is to use pyswarms to find optimal solutions. Now, when I use the option n_processes and run the code in my computer, (Mac with OS Catalina) it runs fine, but when I try to run the same file in a linux computer, it gets stuck and never finishes.
Any ideas of what is wrong and how to fix it?
here is a snippet  of how I am invoking Particle swarms:

```
import numpy as np
import pyswarms as ps

def coarseToPS2(myMat, numOfBestSols, numIterationsPS, numRandomPoints, proRataVac, extraParamsPS, numCores=1):
    [extraParams, myIndex] = extraParamsPS

    # create a list to store the results:
    results = []

    ## NUMBER OF ROWS IN OUTPUT.
    numRows = np.shape(myMat)[0]
    if numRows < numOfBestSols:
        print('numBestSols larger than input matrix')
        sys.exit()

    nm_matrix = np.zeros((numOfBestSols, 6))

    # sort the matrix by the myIndex column.
    sortedMat = myMat[myMat[:, myIndex].argsort()]
    lb = [0] * extraParams[5]
    ub = [1] * extraParams[5]
    bounds = (lb, ub)
    options = {'c1': 0.5, 'c2': 0.3, 'w': 0.9}
    
    kwargs = {"extraParamsPS": extraParamsPS}

    myBestDecisionVariables = np.vstack((sortedMat[:numOfBestSols, 0:5], proRataVac.reshape(np.size(sortedMat[0, 0:5]))))
    
    randomSample = uniformSampleSimplex(dim=extraParams[5], numSamples=numRandomPoints)
    flat1 = myBestDecisionVariables.flatten(order='C')
    flat2 = randomSample.flatten(order='C')
    flat3 = np.append(flat1, flat2)
    numCol = myBestDecisionVariables.shape[1]
    numRow = myBestDecisionVariables.shape[0] + randomSample.shape[0]
    initialSwarm = flat3.reshape(numRow, numCol)
    
    optimizer = ps.single.GlobalBestPSO(n_particles=numOfBestSols + 1 + numRandomPoints,
                                        dimensions=extraParams[5],
                                        options=options,
                                        bounds=bounds,
                                        init_pos=initialSwarm,
                                        n_processes=numCores
                                        )

    output = optimizer.optimize(particleSwarmsObjectiveFunction_General,
                                iters=numIterationsPS, **kwargs)
```

Issue #321, opened 2019-03-21
BinaryPSO does not return best pos


ichbinjakes:
**Describe the bug**
BinaryPSO does not return best pos

**To Reproduce**
Run Feature Subset Selection [notebook](https://github.com/ljvmiranda921/pyswarms/blob/master/examples/feature_subset_selection.ipynb)

After running the optimiser, calling `f_per_particle(pos)` does not return a value equal to `cost`.

**Expected behavior**
`f_per_particle(pos) == cost` should return True

**Environment (please complete the following information):**
 - OS: [Linux]
 - Version [Ubuntu Server 18.04]
 - PySwarms Version [v.1.0.2]
 - Python Version [e.g. 3.6.8]

ichbinjakes:
I am not sure if this is the intended behaviour. The Docs say:
```
def optimize(self, objective_func, iters, **kwargs):
    """ ....
        Returns
        -------
        tuple
            the local best cost and the local best position among the
            swarm.
        """
```
But the code comments say ```# Obtain the final best_cost and the final best_position```. Is  `final_best_pos` the overall best solution found? Is this the same or different to local best position?

The global best pos can be found with:
```
optimizer.swarm.pbest_pos[optimizer.swarm.pbest_cost.argmin(axis=0)]
# or:
optimizer.swarm.best_pos[optimizer.swarm.pbest_cost.argmin(axis=0)]
```

I can submit a PR that fixes the return value or, add a method or variable to access the global best from the BinaryPSO class.

whzup:
Hello @ichbinjakes and welcome to pyswarms! 😄 

Thanks for your report. :+1: It seems like there is a small "typo" in the [code](https://github.com/ljvmiranda921/pyswarms/blob/master/pyswarms/discrete/binary.py#L205). Instead of the `position` argument, it should use the `best_pos` argument as you say. It should return the best position that has been found over the entire course of the optimization. Would be cool to see a PR from you. Could you also fix this bug in the other optimizers?

PR #322, opened 2019-03-23
Fix optimise methods returning incorrect best_pos

ichbinjakes:
## Description of changes
Fix:
- BinaryPSO returning correct `best_pos`
- GeneralPSO returning correct `best_pos`
- LocalBestPSO returning correct `best_pos`
- GlobalBestPSO returning correct `best_pos`

Tests:
- Added simple test cases for each optimiser
  - I have no experience using pytest but tests work, I can add tests using unittest if needed.

Flake8 / Black reformatted some other code

## Related Issue
https://github.com/ljvmiranda921/pyswarms/issues/321

## Motivation and Context
Bug fix

## How Has This Been Tested?
Tested using unittest, then "ported' to the pytest files. All test passed.
Followed instruction in contributing.rst, except used new conda env.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [?] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project. **Tests work but don't match code style**


tests/optimizers/test_general_optimizer.py
```diff
@@ -95,3 +96,14 @@ def test_obj_incorrect_kwargs(self, obj_with_args, optimizer):
         with pytest.raises(TypeError):
             # Wrong kwargs
             cost, pos = optimizer.optimize(obj_with_args, 1000, c=1, d=100)
+
+    def general_eg(self, options):
```
whzup: Instead of adding a function to generate an optimizer you can just add the parameter `optimizer` in the `test_general_correct_pos` method. For a reference see the test methods above. It'll look like this then:
```python
def test_general_correct_pos(self, options, optimizer):
    """ Sanity test to check if the optimizer returns the position which corresponds to the best cost"""
    cost, pos = optimizer.optimize(sphere, iters=5)
    assert sum(pos ** 2) ==  cost
```
You don't need the `_eg` methods then. The same for the other optimizers.
It may be worth to think about these tests a bit more. Maybe there is a more general (non-circular) check we can apply to see if the optimizer really returns the best position found. Something with the optimizer history for example?

Also, please add a docstring under the function header so it has a short description 😄.

pyswarms/backend/handlers.py
```diff
@@ -196,13 +196,15 @@ def reflective(self, position, bounds, **kwargs):
         new_pos = position
         while lower_than_bound[0].size != 0 or greater_than_bound[0].size != 0:
             if lower_than_bound[0].size > 0:
-                new_pos[lower_than_bound] = 2 * lb[lower_than_bound[1]] - \
-                new_pos[lower_than_bound]
+                new_pos[lower_than_bound] = (
```
whzup: Please remove the formatting changes. It is true that the code needs some reformatting but it doesn't belong to a bug fix commit. It also adds a lot of noise in the commit such that it is hard to see what were actual changes.

tests/optimizers/test_binary.py
```diff
@@ -28,3 +29,24 @@ def optimizer_reset(self, options):
         opt.optimize(sphere, 10)
         opt.reset()
         return opt
+
+    def test_binary_correct_pos(self, options):
```
whzup: It may be easier to check whether it found the best by using the [history](https://github.com/ljvmiranda921/pyswarms/blob/master/pyswarms/discrete/binary.py#L183) of the optimizer, what do you think?

ichbinjakes: I think calculating the cost from the position is a good test to have. I also agree with your earlier comment; a need to test that the optimise method indeed returns the best cost? I'll have a look into that.

Thanks for the comments @whzup  :+1:

ichbinjakes: I think I understand what you meant now.

whzup: @ichbinjakes Yes, excuse me, I may have not expressed myself clearly enough 😕:smile:

whzup:
This seems good to me :+1: Thank you for your PR @ichbinjakes. Hope you had fun working on it 😄🎊

ljvmiranda921:
Can you squash-merge this, @whzup ? Thank you @ichbinjakes !

ljvmiranda921:
Nice!! Thank you again, @ichbinjakes , and to @whzup for conducting the review! I'll be updating the contributor's list this weekend!

ljvmiranda921:
@allcontributors[bot] please add @ichbinjakes for bug and code

PR #370, opened 2020-02-11
Fix memory issues when running optmize in a loop

ddrm86:
## Types of changes

## Checklist:


whzup:
Perfect @ddrm86! Thank you so much for your contribution! So do you know why this occurs with multiprocessing? Because to me, it seems like the multiprocessing should only affect the objective function and not the entire `optimize` function, but I don't have any experience with multiprocessing 😄.

ddrm86:
You are welcome! I have no clue about multiprocessing, but if the optimize function keeps making the history larger

`hist = self.ToHistory(`

without never cleaning it, it does not matter it it was called in a sequence or in parallel.

Issue #374, opened 2020-03-27
Citing PySwarms in my work


Timmenke:
Hi PySwarms team, I've extensively used your package in my work on automatic design of quantum computing hardware: https://arxiv.org/abs/1912.03322. If you like, please add my preprint to the readme as work that is citing you. Great package, I love it!

whzup:
@Timmenke Thanks a lot! We'll gladly add you as soon as we get around to it 😄 The paper looks very interesting by the way!

ljvmiranda921:
Hi @whzup and @Timmenke , let me take point on updating the README. :) Unless @Timmenke wants to open up a Pull Request?

Timmenke:
Thanks a lot for including my paper in the update!

PR #380, opened 2020-04-04
Update references with new citations (#374)

ljvmiranda921:
Updates the references from last year :)


ljvmiranda921:
Yeah I was pretty surprised as well! :O

PR #386, opened 2020-04-19
Minor fixes

ckastner:
This resolves a rendering issue with [handlers](https://pyswarms.readthedocs.io/en/latest/api/pyswarms.handlers.html) (see, for example, [here](https://pyswarms.readthedocs.io/en/latest/api/pyswarms.handlers.html#pyswarms.backend.handlers.BoundaryHandler.nearest)), as well as minor docstring issues producing warnings during documentation rebuild.


ljvmiranda921:
Hi @ckastner thanks for this! @all-contributors please add @ckastner for docs and platform

Issue #389, opened 2020-04-20
Revisit and update documentation and examples


ljvmiranda921:
I think one of the more common issues we encounter is that some examples don't work out of the box. We may have introduced a lot of changes in the API from the previous versions and weren't able to update the docs properly. 

Let's scope this one out, right now we have docs located in:
- Our docstrings
- Jupyter notebooks

I know we did this exercise before, but it may be good to revisit it again just to be really sure that everything is working well.

whzup:
I found some unrendered math [here](https://pyswarms.readthedocs.io/en/latest/examples/usecases/electric_circuit_problem.html).

whzup:
I also think we should open an issue for a collection of architecture updates we want to make (e.g. see #235, #259).

ljvmiranda921:
^Agree! Or we can just reuse the existing issues #235 #259 and include them in this current effort.
Also, no pressure here, I appreciate that you're taking the time and holding the fort answering Issues and reviewing PRs, but focus on your thesis and graduation! I'm rooting for you!

msat59:
> I think one of the more common issues we encounter is that some examples don't work out of the box. We may have introduced a lot of changes in the API from the previous versions and weren't able to update the docs properly.
> 
> Let's scope this one out, right now we have docs located in:
> 
> * Our docstrings
> * Jupyter notebooks
> 
> I know we did this exercise before, but it may be good to revisit it again just to be really sure that everything is working well.
> 
> Let me know what you think @whzup @stevenbw

In case of a n_process>2, an addition consideration must be given to Jupyter Notebook for windows users.

https://stackoverflow.com/questions/47313732/jupyter-notebook-never-finishes-processing-using-multiprocessing-python-3

whzup:
More unrendered math https://pyswarms.readthedocs.io/en/latest/api/pyswarms.backend.html#module-pyswarms.backend.handlers

whzup:
@ljvmiranda921 Do you think it would be a good idea to build a MWE for a custom optimizer in [Writing your own optimization loop](https://pyswarms.readthedocs.io/en/latest/dev.loop.html)?

ljvmiranda921:
^Ooooh, you meant like a scaffold of it? @whzup yeah that would be good to be honest. I think just describing what it looks like (what I did) isn't enough

svhoy:
Hello, 

[...]

PR #399, opened 2020-05-05
Fix unrendered math

whzup:
## Description

Fixed some math that was badly rendered in the documentation and examples.

## Related Issue

#389

## Motivation and Context

We want the documentation to stay up to date and clean.

## Types of changes

## Checklist:


stevenbw:
Thanks for kicking this off 😄 

Is this PR just for the Tex parts that haven't rendered? Might be worth splitting #389 across multiple PRs.

whzup:
Yes, currently only the TeX parts 😄 Would be a good idea to split it up, but we need something like a TODO list first to see what belongs where 😕

ljvmiranda921:
Hey @whzup and @stevenbw , is this good for review? There's still a WIP tag so I'm assuming this is still work-in-progress.

I agree that we should split this to multiple PRs. I create one big Issue and pinned it down so that it can be the focus of our work as well. Let me update that Issue to specify which one fits into what so that we can separate these tasks

Issue #427, opened 2020-10-31
How to change Hyperparameters on runtime?


Antonio-Nappi:
* pywarm = 1.1.0
* Python version: 3.7.8
* Operating System: Windows 10

### Description
Hi everybody, I'm new in pyswarm and I'm still studying its architecture but maybe you could help me to find the right answers. I would like to implement a PSO with TVAC (Time-Varying Acceleration Coefficients) and these coefficents depends on the current iteration and the max_iteration number. Looking in the file I've seen that the velocity and position update are made in operators script. So my question is: should I define a new method in that script (maybe a "me_velocity_update") with the right coefficients and then changing the behaviour of the topology classes in order to let them use my new method? In this way how could I pass the current iteration number? Or (maybe) there is a simpler way than this which I've not considered?

ljvmiranda921:
Hi @Antonio-Nappi , can you check the merged Pull Request above if this answers your request? 
Right now, I don't see any possible way to update coefficients during runtime except for the newly-implemented handlers. Perhaps the best way to do PSO with TVAC is to implement your own handler (following the PR above), and pass it during the init stage

Antonio-Nappi:
Hi @ljvmiranda921, I think that with linear decay I could fix my problem. I've seen that it's an option handler and now I'm just looking at the doc to understand how to use it, but I don't find and example. Could you provide it to me? Could you also say me if it's possible to use an option handler with the GlobalBestPSO or should I make subclasses from yours?

ljvmiranda921:
Hi @Antonio-Nappi , can you try something like this?

```python
import pyswarms as ps
from pyswarms.utils.functions import single_obj as fx
options = {'c1': 0.5, 'c2': 0.3, 'w':0.9}
optimizer = ps.single.GlobalBestPSO(
    n_particles=10, 
    dimensions=2, 
    oh_strategy="lin_decay",   # Set it from here
   options=options
)
best_cost, best_pos = optimizer.optimize(fx.sphere, iters=100)
```

Antonio-Nappi:
Yes, I have the latest version (downloaded specifying the version inside pycharm) but I got the following error 
`TypeError: __init__() got an unexpected keyword argument 'oh_strategy'`
and can you also say me how it's possible to set boundaries for coeffients?

ljvmiranda921:
[...]

Issue #423, opened 2020-10-01
How to vary inertia during optimisation?


Ron-Leizrowice:
Hi, new to using Pyswarms, I'm wondering if there is a way to implement variable inertia optimisation using the "GeneralOptimizerPSO" module? I want to decrease the 'w' parameter in options as my optimisation progresses, but I don't know if updating the global variable of 'options' actually changes the optimiser hyperparameters?

ljvmiranda921:
Hi @Ron-Leizrowice , yup updating the global variable options should change the optimizer hyperparameters :smile:

Ron-Leizrowice:
Amazing thank you!

Antonio-Nappi:
could you post a code snippet to show how to change hyperparameters according to the current iterarion?

PR #429, opened 2020-11-05
Added handler for options

nikhil-sethi:
## Description
This pull request adds an options handler which enables the use of different update strategies for varying the options dict during the optimization run. The current version has only the following strategies but more can be added easily:
- Exponentially decay
- Linear decay
- Random
- Non-linear modulation index
A lot of unchanged files reformatted after running black

## Related Issue
-https://github.com/ljvmiranda921/pyswarms/issues/427
-https://github.com/ljvmiranda921/pyswarms/issues/423

## Motivation and Context
Varying the options/coefficients during runtime is a well-accepted technique to balance exploration vs exploitation tradeoff which improves convergence significantly.

## How Has This Been Tested?
Added basic tests in test_handler
Manually tested on Python 3.7.4, Windows 10

## Types of changes

## Checklist:


ljvmiranda921:
Hi @nikhil-sethi , thanks for the PR! Let me review this next week! This seems to be a big PR so let me take a look for a while :)

pyswarms/backend/handlers.py
```diff
@@ -540,3 +542,193 @@ def zero(self, velocity, clamp=None, **kwargs):
...
+        that is passed at initialization to account for time-varying coefficients. An example
+        for the usage:
+
+        .. code-block :: python
+
+            from pyswarms.backend import operators as op
+            from pyswarms.backend.handlers import OptionsHandler
+
+
+            oh = OptionsHandler(strategy=(None, None, "exp_decay"))
```
ljvmiranda921, line 577: Are we expecting the user to use multiple strategies at the same time? Or do we just expect a single one every time?


1. If multiple, then maybe passing a list, without explicitly stating `None`s should be better:

```python
# If expecting multiple strats
oh = OptionsHandler(strategy=["exp_decay", "lin_decay"])  # will just apply exp_decay and lin_decay

# If single strat only
oh = OptionsHandler(strategy=["exp_decay"])
```

Then within your code you can just iterate like so:

```python
# Assuming: strategies =  {"exp_decay": exp_decay, "lin_decay": lin_decay, "some_decay": some_decay}
choices = ["exp_decay", "lin_decay"]

for choice in choices:
    strategy = strategies[choice]
    strategy()
```

2. If we're just expecting a single strategy from the user every time, then I think passing a string is way better

```python
# If single strat only
oh = OptionsHandler(strategy=["exp_decay"])
```

ljvmiranda921 (commented):
Just a few design comments! Thanks for this PR, this seems to be very helpful!

pyswarms/backend/handlers.py
nikhil-sethi: Hi @ljvmiranda921! Thanks for the review. Would love to contribute more.

We definitely want to allow different strategies for different options. This might be useful in something like a Time-varying acceleration PSO. 
**Why use a tuple not a list:**
1(a) A list could surely be used without any change but I purposely used a tuple as part of an attempt to avoid mutable arguments. Although the strategy iterable is not getting mutated within any function, it was just to avoid any confusion and is just good python practice.

**Why use an explicit None**
1(b) Say the user wants to do the following: 
Nonlinear modulation decay for the cognitive component (c1)
Nothing(constant) for the social component (c2)
Exponential decay for the inertia weight(w)

```python3
options = {"c1": 0.5, "c2": 0.3, "w": 0.9}   # will be used as the start options 

# A if we dont have explicit None`
oh_strategy=("nonlin_mod", "exp_decay")

# B if we have explicit None
oh_strategy=("nonlin_mod", None,  "exp_decay")

# Call instance of PSO
optimizer = ps.single.GeneralOptimizerPSO(
    n_particles=10,
    dimensions=2,
    topology=Star(),
    options=options,
    oh_strategy=oh_strategy,
)
cost, pos = optimizer.optimize(fx.sphere, iters=100)

# inside the call of options handler it is something similar to 
for key, strat in zip(options.keys(), oh_strategy):
    if strat is None:
        continue
    "change the option with "key" to value corresponding to "strat"
```
Since we have to loop over the strategy tuple, the first option will take up the first strategy and so forth.
This means that with method A we will result in an exponential decay for the second option i.e. "c2". Which is wrong because we need it to be constant.
This need for correct order is definitely a small weakness in some sense but the alternative solution was to use a dict (or a namedTuple maybe?) for the oh_strategy which wasn't very consistent with the other handlers and is also mutable again.

2. Agreed. If the same option for all coefficients was better in terms of functionality, a string would be best.

Let me know what you think! I dont mind changing anything and everything as long as its better in terms of style, convention, speed etc.

ljvmiranda921: > Since we have to loop over the strategy tuple, the first option will take up the first strategy and so forth.
This means that with method A we will result in an exponential decay for the second option i.e. "c2". Which is wrong because we need it to be constant.

Hi @nikhil-sethi now makes sense why you opted for this data structure! Because it applies to c1, c2, and w. Hmmm, perhaps it's fine to make it explicit rather than implying based on index positions (first position is c1, second position is c2, third position is w). This then allows more readability on the user side and reduces any surprises in the long run! 

 > This need for correct order is definitely a small weakness in some sense but the alternative solution was to use a dict (or a namedTuple maybe?) for the oh_strategy which wasn't very consistent with the other handlers and is also mutable again.

I'm fine using a dictionary for this, even if it's not consistent with the other handlers. I guess you just need to override some of the inherited methods so that it accepts your data structure. How about this one?

```python
options = {"c1": 0.5, "c2": 0.3, "w": 0.9}   # will be used as the start options 
# Multiple options
oh_strategy={"w":"exp_decay", "c1":"nonlin_mod"}
# Single option
oh_strategy={"c1":"nonlin_mod"}
``` 

Let me know what you think!

nikhil-sethi: Okay. I agree it's more readable and intuitive. I didn't want to do it before without discussing it as it would have broken style.  I'll make changes as soon as I can :)

ljvmiranda921: @nikhil-sethi Looking forward to it!

ljvmiranda921:
Saw it was updated! Will review tomorrow! Thanks @nikhil-sethi :+1:

ljvmiranda921 (approved):
LGTM!

ljvmiranda921:
Hey thanks for this wonderful feature! It's good that we're able to extend the handlers effectively, and created a design that allows finer control for such states! If you wish to add more into this, you can add a Jupyter notebook documentation that demonstrates particle changes for each handler, so that this feature may be more visible to the public!

@all-contributors please add @nikhil-sethi for code and documentation!

ljvmiranda921:
@all-contributors please add @nikhil-sethi for code and docs

nikhil-sethi:
> Hey thanks for this wonderful feature! It's good that we're able to extend the handlers effectively, and created a design that allows finer control for such states! If you wish to add more into this, you can add a Jupyter notebook documentation that demonstrates particle changes for each handler, so that this feature may be more visible to the public!

Thanks for the merge and review! I think the latest commit after the review is much better. I'll try and get around to doing the Jupyter doc sometime in the near future. 👍

Antonio-Nappi:
Thanks @nikhil-sethi for this feature, but could you provide a more exhaustive example? I think that the one present in the doc is a bit skimpy.

nikhil-sethi:
Hi @Antonio-Nappi. I wanted to make a good Jupyter notebook for this but haven't been able to get much time off of work yet. Will try and do it this week itself.

nikhil-sethi:
@ljvmiranda921. This reminds me. There was a potential issue that I thought might need some discussion or just some understanding on my part. 
Say I want to use the `invert` strategy for the `VelocityHandler`. There is a `z (shrinking factor)` variable that is required for this. The default value for this is 0.5 which works fine but say I want to change this. This can be done via the call to `vh` under `compute_velocity` function in `operators.py` like so (Although we will need to pass and unpack kwargs in this call which doesnt exist right now):
```
# Compute temp velocity (subject to clamping if possible)
temp_velocity = (w * swarm.velocity) + cognitive + social
updated_velocity = vh (temp_velocity, clamp, position=swarm.position, bounds=bounds )
```
This can be done fairly easily if the user creates their own optimization loop and then continously passes `z` through the respective velocity functions but this can't be done (I think) through the outer `optimize` call as it only accepts the keyword arguments for the objective function. 
I ask this because in the options handler case, it will be common for the user to set their own end options. Currently, one has to write their own loop for this. The `optimize` call only allows different strategies but default ending options. 
Any thoughts regarding this? I might be misinterpreting everything

ljvmiranda921:
Hi @nikhil-sethi yeah this makes sense. Am I reading it right that there should be a way to atleast handle those options without writing the optimization loop? Because as far as I can tell if they want pure customization, they really need to write their own loop. But there are some common settings that is not exposed on the core optimizers..

Would you mind opening another Issue and discuss it there? I'm starting to think that maybe instead of passing the parameters themselves, we pass instances of the handlers... then we make the handlers generic enough so that it works with whatever implementation.. it's a big refactor though. But a crude example may be like this:

```python
myhandler = VelocityHandler(**my_params)
optimizer = GlobalBestPSO(..., v_handler=VelocityHandler(**my_params), some_handler=SomeHandler(**my_params).
```

Then the default values can be `None` or any default that is sensible. In that way, everything is controlled outside...
Hmm, let's see first. I guess we need a solution that works with the current design, but could be nice if we can try a short refactor of it. It's a big task though, and we all have day jobs to do :sweat_smile: . I think let's discuss potential solutions for the current design first

nikhil-sethi:
Yes you're reading it right. I agree that if an above average user might write their own loop anyway but it's a "good to have" sort of thing.
Yes, I'll open another issue to discuss everything including the possibility itself.

Issue #439, opened 2020-12-10
Troubling in writing my own optimization loop


Antonio-Nappi:
* PySwarms version: 1.2
* Python version: 3.7
* Operating System: Windows 10

### Description
I want to implement my own optimization loop in order to manage coefficients in runtime. I followed the istructions present in the doc. The problem is raised by scipy, but if I use the pyswarms' optimizer this problem is not present. Why?

### What I Did
This is my code:
```
c1i = 2.5
c1f = .5
c2i = .5
c2f = 2.5
iterations = 100
my_topology = Star()  # The Topology Class
options = {'c1': c1i, 'c2': c2i, "w": 0.7298}
lower_bound = np.asarray([0 for _ in range(9)])
upper_bound = np.asarray([1, 1200, 1, 1, 1, 1, 1, 1, 1])
for i in range(iterations):
            print("Iteration={},best cost={}".format(i + 1, my_swarm.best_cost))

            my_swarm.current_cost = cost_f(my_swarm.position, gt, times, population,
                                           total_active)  # Compute current cost

            my_swarm.pbest_cost = cost_f(my_swarm.pbest_pos, gt, times, population,
                                         total_active)  # Compute personal best pos
            my_swarm.pbest_pos, my_swarm.pbest_cost = P.compute_pbest(my_swarm)  # Update and store

            # Part 2: Update global best
            # Note that gbest computation is dependent on your topology
            if np.min(my_swarm.pbest_cost) < my_swarm.best_cost:
                my_swarm.best_pos, my_swarm.best_cost = my_topology.compute_gbest(my_swarm)

            my_swarm.velocity = my_topology.compute_velocity(my_swarm)
            my_swarm.position = my_topology.compute_position(my_swarm)
            my_swarm.options["c1"] = (c1f - c1i) * (i + 1) / iterations + c1i
            my_swarm.options["c2"] = (c2f - c2i) * (i + 1) / iterations + c2i
```
As you can see, I've even left the original comments of the code.
The problem appears at the fourth iteration, and it is the following:
```
ODEintWarning: Excess work done on this call (perhaps wrong Dfun type). Run with full_output = 1 to get quantitative information.
 warnings.warn(warning_msg, ODEintWarning)
```

PR #440, opened 2020-12-11
Options handler example + fixes

nikhil-sethi:
## Description

- Added a notebook about handling options under docs/examples/tutorials
- fixed a bug in linear strategy
- linear strategy will work for both growth and decay- hence renamed.
- removed unnecessary '.keys()' from some of 'kwargs["end_opts"]"

## Related Issue
#439 (indirect solution)

## Motivation and Context
Helps in understanding the options handler better

## How Has This Been Tested?
Manually tested
test_handlers.py reflected with changes

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:


ljvmiranda921:
Thank you so much for this @nikhil-sethi ! Will review this weekend :+1:

ljvmiranda921 (approved):

pyswarms/backend/handlers.py
```diff
@@ -574,11 +574,11 @@ def __init__(self, strategy):
             from pyswarms.backend.handlers import OptionsHandler
 
 
-            oh = OptionsHandler(strategy=(None, None, "exp_decay"))
+            oh = OptionsHandler(strategy={ "w":'exp_decay', "c1":'nonlin_mod',"c2":'lin_variation'})
```
ljvmiranda921, line 577: @nikhil-sethi nit: just run `black` here to convert single-quotes to double-quotes

pyswarms/backend/handlers.py
```diff
@@ -600,8 +600,8 @@ def __call__(self, start_opts, **kwargs):
             if not self.strategy:
                 return start_opts
             return_opts = copy(start_opts)
-            for key in start_opts.keys():
-                if key in self.strategy.keys():
+            for key in start_opts:
```
ljvmiranda921, line 603: Maybe there's a more specific variable name than `key`? If we're iterating across a parameter or a config I think we can use something like`cfg` etc.

ljvmiranda921 (approved):

pyswarms/backend/handlers.py
nikhil-sethi: Sorry for the late reply. How about `opt`/`options` since we're iterating over the keys in `start_opts`, `end_opts`, etc.?

ljvmiranda921: That works! Any is fine :)

pyswarms/backend/handlers.py
nikhil-sethi: Sure. Will do it.

Issue #447, opened 2021-02-07
wrong implementation/definition of ftol_iter


D4T4-80Y:
**Description**

When **ftol** is used, for example 1e-10, pyswarm stops when `swarm.best_cost` is lower than `ftol`  for `ftol_iter` times. However, the correct naming of `ftol_iter` should be `stall_iter`.  This means that the search stops when best_cost isn't improved after x iterations, or when it is below `ftol`. However, when I search for a solution, I need to stop the search when the best_cost isn't improved after `x` iterations or when it is below `ftol` (must happen only once).

The current naming and implementation is confusing and doesn't make sense. That's why I thought I had to set `ftol_iter=1` which caused stopping after the first iteration and having premature solution ([0.03840509 0.00329153] instead of [-9 -9]).

In this regard, the  default value of ftol_iter should be big enough (at least 100) to avoid having premature solution. ftol_iter=1 is unusual and very small. 

This is what I got with ftol=1e-10 and ftol_iter=100. As can be seen, it must stop after iteration 199.

```
196 = {float64} 1.543800321250907e-10
197 = {float64} 1.543800321250907e-10
198 = {float64} 1.543800321250907e-10
199 = {float64} 1.2788497031507338e-12   <-- stop here
200 = {float64} 1.2788497031507338e-12
201 = {float64} 1.2788497031507338e-12
202 = {float64} 1.2788497031507338e-12
203 = {float64} 1.2788497031507338e-12
204 = {float64} 1.2788497031507338e-12
```

**Sample code**
A sample code is attached. The function to be optimized is a modified sphere function centered at [-9,-9]. The real cost is **zero**.

[ftol_test.zip](https://github.com/ljvmiranda921/pyswarms/files/5940520/ftol_test.zip)

**Expected behavior**
As I said, I expect that the optimizer stops after `ftol_iter` iterations when there is no improvement (best_cost stalls) or it stops **instantly** when the best_cost is below `ftol`. With the current implementation, optimizer only stops after iteration around 300, after stalling for 100 iterations but with ftol around 2e-16, which is much lower than given ftol. However, it must stop somewhere around iterations 200.

**Environment (please complete the following information):**
 - PySwarms Version 1.3.0

Thank you.
DATA-BOY

D4T4-80Y:
[...]

PR #449, opened 2021-02-19
Improving ftol_iter

msat59:
The descriptions of `ftol` and `ftol_iter` were improved. Also, the default value of `ftol_iter` was changed from 1 to 50, to avoid a premature solution. In addition, the code of `ftol_iter` was modified to make it more readable. Finally, a default value of 1000 was defined for `iters` in **optimize** method. So it can be used simply by calling `optimizer.optimize(obj_func)`.

## Description
As discussed here [(#447)](https://github.com/ljvmiranda921/pyswarms/issues/447), there was confusion in the description and purpose of using `ftol_iter`. By improving `ftol` and `ftol_iter` descriptions, and changing `ftol_iter` default value, one can use it as the `stall_iter` feature without any issue with a premature solution. 

## Motivation and Context
Confusion in understanding the purpose of ftol_iter, and its default value was resulting in a premature solution.

## Types of changes

- [ x] Updating parameters descriptions.
- [ x] Updating parameters default values.

## Checklist:

### DIFF
--- a/pyswarms/base/base_discrete.py
+++ b/pyswarms/base/base_discrete.py
@@ -49,7 +49,7 @@ class DiscreteSwarmOptimizer(abc.ABC):
         velocity_clamp=None,
         init_pos=None,
         ftol=-np.inf,
-        ftol_iter=1,
+        ftol_iter=50,
     ):
         """Initialize the swarm.
 
@@ -81,13 +81,16 @@ class DiscreteSwarmOptimizer(abc.ABC):
             a tuple of size 2 where the first entry is the minimum velocity
             and the second entry is the maximum velocity. It
             sets the limits for velocity clamping.
-        ftol : float, optional
-            relative error in objective_func(best_pos) acceptable for
-            convergence. Default is :code:`-np.inf`.
+        ftol : float
+            relative error in objective_func(best_pos) acceptable for convergence.
+            To deactivate it in order to process all iterations, use any negative value.
+            Deactivating ftol also disables ftol_iter.
+            Default is :code: `-np.inf` to disable ftol
         ftol_iter : int
-            number of iterations over which the relative error in
-            objective_func(best_pos) is acceptable for convergence.
-            Default is :code:`1`
+            number of consecutive iterations over which the relative change in
+            objective_func(best_pos) is stalled or less than ftol. It works
+            when ftol is greater than zero (e.g. 1e-6)
+            Default is :code:`50`
         options: dict
             a dictionary containing the parameters for a specific
             optimization technique
@@ -146,7 +149,7 @@ class DiscreteSwarmOptimizer(abc.ABC):
         self.velocity_history.append(hist.velocity)
 
     @abc.abstractmethod
-    def optimize(self, objective_func, iters, n_processes=None, **kwargs):
+    def optimize(self, objective_func, iters=1000, n_processes=None, **kwargs):
         """Optimize the swarm for a number of iterations
 
         Performs the optimization to evaluate the objective
@@ -158,7 +161,7 @@ class DiscreteSwarmOptimizer(abc.ABC):
         objective_func : callable
             objective function to be evaluated
         iters : int
-            number of iterations
+            number of iterations. Default is :code:`1000`
         n_processes : int
             number of processes to use for parallel particle evaluation
             Default is None with no parallelization
--- a/pyswarms/base/base_single.py
+++ b/pyswarms/base/base_single.py
@@ -50,7 +50,7 @@ class SwarmOptimizer(abc.ABC):
         velocity_clamp=None,
         center=1.0,
         ftol=-np.inf,
-        ftol_iter=1,
+        ftol_iter=50,
         init_pos=None,
     ):
         """Initialize the swarm
@@ -82,13 +82,16 @@ class SwarmOptimizer(abc.ABC):
             sets the limits for velocity clamping.
         center : list, optional
             an array of size :code:`dimensions`
-        ftol : float, optional
-            relative error in objective_func(best_pos) acceptable for
-            convergence. Default is :code:`-np.inf`.
+        ftol : float
+            relative error in objective_func(best_pos) acceptable for convergence.
+            To deactivate it in order to process all iterations, use any negative value.
+            Deactivating ftol also disables ftol_iter.
+            Default is :code: `-np.inf` to disable ftol
         ftol_iter : int
-            number of iterations over which the relative error in
-            objective_func(best_pos) is acceptable for convergence.
-            Default is :code:`1`
+            number of consecutive iterations over which the relative change in
+            objective_func(best_pos) is stalled or less than ftol. It works
+            when ftol is greater than zero (e.g. 1e-6)
+            Default is :code:`50`
         """
         # Initialize primary swarm attributes
         self.n_particles = n_particles
@@ -144,7 +147,7 @@ class SwarmOptimizer(abc.ABC):
         self.velocity_history.append(hist.velocity)
 
     @abc.abstractmethod
-    def optimize(self, objective_func, iters, n_processes=None, **kwargs):
+    def optimize(self, objective_func, iters=1000, n_processes=None, **kwargs):
         """Optimize the swarm for a number of iterations
 
         Performs the optimization to evaluate the objective
@@ -156,7 +159,7 @@ class SwarmOptimizer(abc.ABC):
         objective_func : function
             objective function to be evaluated
         iters : int
-            number of iterations
+            number of iterations. Default is :code:`1000`
         n_processes : int
             number of processes to use for parallel particle evaluation
             Default is None with no parallelization
--- a/pyswarms/discrete/binary.py
+++ b/pyswarms/discrete/binary.py
@@ -77,7 +77,7 @@ class BinaryPSO(DiscreteSwarmOptimizer):
         velocity_clamp=None,
         vh_strategy="unmodified",
         ftol=-np.inf,
-        ftol_iter=1,
+        ftol_iter=50,
     ):
         """Initialize the swarm
 
@@ -114,12 +114,15 @@ class BinaryPSO(DiscreteSwarmOptimizer):
             a strategy for the handling of the velocity of out-of-bounds particles.
             Only the "unmodified" and the "adjust" strategies are allowed.
         ftol : float
-            relative error in objective_func(best_pos) acceptable for
-            convergence
+            relative error in objective_func(best_pos) acceptable for convergence.
+            To deactivate it in order to process all iterations, use any negative value.
+            Deactivating ftol also disables ftol_iter.
+            Default is :code: `-np.inf` to disable ftol
         ftol_iter : int
-            number of iterations over which the relative error in
-            objective_func(best_pos) is acceptable for convergence.
-            Default is :code:`1`
+            number of consecutive iterations over which the relative change in
+            objective_func(best_pos) is stalled or less than ftol. It works
+            when ftol is greater than zero (e.g. 1e-6)
+            Default is :code:`50`
         """
         # Initialize logger
         self.rep = Reporter(logger=logging.getLogger(__name__))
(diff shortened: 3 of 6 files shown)


ljvmiranda921:
@msat59 , thanks for the PR! Let's first ensure that the tests pass (e.g. maybe update some of the parameters being called, or change the expected results given the new implementation).

Also hoping if possible for @nishnash54 to checkin on this and give his review (if he has time)

pyswarms/base/base_discrete.py
```diff
@@ -81,13 +81,16 @@ def __init__(
...
+            Deactivating ftol also disables ftol_iter.
+            Default is :code: `-np.inf` to disable ftol
         ftol_iter : int
-            number of iterations over which the relative error in
-            objective_func(best_pos) is acceptable for convergence.
-            Default is :code:`1`
+            number of consecutive iterations over which the relative change in
+            objective_func(best_pos) is stalled or less than ftol. It works
+            when ftol is greater than zero (e.g. 1e-6)
+            Default is :code:`50`
```
nishnash54, line 93: I'm not exactly convinced how `ftol_iter` can work for both the **acceptable tolerance** and **stall** cases. Maybe I'm missing something here.

The stall condition is when there is no improvement in the optimal value over `n_iter_no_change` (`stall_iter`) iterations. The `ftol_iter` should be checking that the `ftol` threshold must be achieved for the `ftol_iter` consecutive iterations.

Ill try to draw up an example but just for a sanity check (minimization problem)
```py
.
19: 50
20: 50    # <- stalled for 20 iterations (stop here)
21: 50    # <- if no stall iter, keep going as ftol is not achieved (will run all the iterations)
.
```
@ljvmiranda921 Are we on the same page here?

pyswarms/base/base_discrete.py
```diff
@@ -49,7 +49,7 @@ def __init__(
         velocity_clamp=None,
         init_pos=None,
         ftol=-np.inf,
-        ftol_iter=1,
+        ftol_iter=50,
```
nishnash54, line 52: Also I don't think this is a good change. `ftol_iter`defaults to `1` to preserve backward compatibility with the initial implementation of `ftol`. This is the most basic case of early stopping when we stop as soon as an accepted tolerance is achieved.

Increasing the default value goes against the purpose. I think that the default should be 1 and the user should change the value as per their usecase.

nishnash54 (changes requested):
I've just had a brief look at the changes. Once the tests pass, I'll spend some more time on it.

@ljvmiranda921 Maybe we should first clearly draw out the functionality we are trying to implement here.

pyswarms/base/base_discrete.py
msat59: This is what I thought before. All reasons for confusions are the naming of ftol_iter.

The current implementation is what I exactly was trying to implement, but with different default values. If you like to keep `ftol_iter=1` and I implement something new, we are using the same algorithm twice.

pyswarms/base/base_discrete.py
msat59: @nishnash54 , you are missing what I was missing. With the current implementation, you never reach to iteration #21 with ftol_iter=1. You are stopped after 1st or 2nd iteration.

Please check the next post.

pyswarms/base/base_discrete.py
nishnash54: The case that you are talking about is actually pretty common (when er are using random swarm initialization)
> There is one case that delta < ftol may happen at 1st or 2nd iteration, and it's when n_particles is huge, something like 10,000.

 You can test the same by checking  out `tests/optimizers/test_tolerance.py` Just run the `GlobalBestPSO`, make it fail (`assert False`) and update the code to print the delta at each iteration.

I've generated the below sample output the same way (running the current tests) and added a `print(i, delta)` to the `global_best.py` file.

1. `test_ftol_effect (ftol = 0.01)`
```py
0 False
1 False
2 False
3 True    # <- stops immediately
```
2. `test_ftol_iter_effect (ftol = 0.01, ftol_iter=50)`
```py
0 False
1 True    # <- would have stopped here (if ftol_iter=1)
2 True
3 False
4 False
5 True
6 True
7 False
8 True
9 True
10 True
11 True
12 True
...
57 True    # <- finally stops at iteration 57 (after 50 acceptable iterations)
```
Additionally, in the library delta is not defined as the difference between two consecutive iterations. It is the difference between the `current_best_cost` and the `global_best_cost`. So for the `stall_iter` (`n_iter_no_change`) condition, we need to check the improvements in the `global_best_cost` (not delta)

```py
# delta calculation
relative_measure = self.ftol * (1 + np.abs(best_cost_yet_found))
delta = (
    np.abs(self.swarm.best_cost - best_cost_yet_found)
    < relative_measure
)
```

@msat59 Let me know if you have any issues replicating these test results. @ljvmiranda921 Please do point out if I missed anything.

msat59: @nishnash54, this is what I saw last night in **Tolerance** jupyter notebooks in docs\examples\tutorials folder.

`ftol_iter` helps to generalize the ftol property and extends the feature over 'n' iterations. As a fixed tolerance might seem too restrictive, ftol_iter provides the optimizer some breathing space to improve the best cost. If there are no improvements detected over ftol_iter subsequent iterations, the optimizer stops.

msat59: I hear what you say @nishnash54. but...

when there is no improvement (stalled), the `best_cost_yet_found` is actually `self.swarm.best_cost`. so delta gets zero, which is less than `relative_measure`.

Here is the results at 4th iteration, in my test case.

There is no improvement in last two iterations.
image

and the checks are
image
image

You see... `self.swarm.best_cost = best_cost_yet_found` so **delta is True** because np.abs(0.023167180458767107 - 0.023167180458767107) = 0 < 1.0231671804587672e-06.

so deque([False, False, **True**, **True**], maxlen=50)

nishnash54: The numbers you have presented are correct. That is the actual cost history of the swarm over the iterations. But `ftol` does not deal with the actual costs. The cost based early stopping is not yet implemented. `ftol` is a relative measure.

The stopping criterion is defined by the improvement on the `best_cost_yet_found`. So if the change is big enough the iterations continue.

```py
# print(i, self.swarm.best_cost, best_cost_yet_found, delta)
22 -56.0 -56.0 True
23 -56.0 -56.0 True
24 -56.0 -56.0 True
25 -58.0 -56.0 False  # <- Here the change is big enough to let the algorithm continue
26 -58.0 -58.0 True
27 -58.001 -58.0 True  # <- Here there is a small improvement (not big enough for delta to change)
                       # This is the exact case where we still want to consider the iterations for
                       # early_stopping, but on the other hand the algorithm has not stalled.
28 -58.001 -58.001 True
29 -58.001 -58.001 True
```
Again this implementation is early_stopping based and not based on a cutoff threshold (as that will change for each optimization problem). If you're having problems running the tests cases, please try running through the logic with some sample values in a list.

> I think it is better to withdraw this PR and submit one with this implementation to have both stall_iter and ftol_iter. I didn't want to make things complicated but it seems there is no other choice.

@ljvmiranda921 Let us know how you want to proceed with this.

msat59: > The numbers you have presented are correct. That is the actual cost history of the swarm over the iterations. But ftol does not deal with the actual costs. The cost based early stopping is not yet implemented. ftol is a relative measure.

@nishnash54, I know that ftol deals with differences in costs and relative measure. I took those screenshots to show you that the current implementation deals with the difference between **two consecutive iterations** which counts stalled iterations.

If I implement it in my own way, I have to change your implementation to exclude stalled iterations. I'm sure you complain later that it's not working as you have implemented ftol_iter. 

Guys, I think we have a conflict of interest here. :wink:
I feel you are against any changes in your implementation, which has resulted in hundreds of lines of explanations and then may cause the introduction of complexity in PySwarms functionality.

I'm happy with your implementation, however, it is not 100% efficient computationally. If the iters=10000 and ftol_iter is 10000 for instance, an array of 10k boolean values must be processed 10k times (all(ftol_history), which is not an efficient way; but I didn't complain. I requested for renaming `ftol_iter` to `stall_iter` but you were not happy. Then I kept whatever we had and just changed the default values and then you were not happy.

Please tell me what I should do. I'm really stalled. :laughing: 
If I forget about any changes I made now and withdraw the PR, and implement **fmin** feature, I need a `stall_iter` parameter, which is exactly what you've named it as `ftol_iter`. And it's default value must be greater than 1 to avoid premature optimization, which is what ftol_iter does at the moment.

**Please explain why should we have a duplicated parameter?**

If I implement my own way and keep ftol_iter as it is (there will be some conflicts), it is like having a car, using its boot as a storage container, and wanting to buy a new and exactly the same car because we want to drive to work.

That's the story guys. Please advise.
@nishnash54 
@ljvmiranda921

D4T4-80Y:
This implementation is almost what has been available in Matlab PSO toolbox, and looks fine. As @msat59  explained, this implementation is less expensive than all(ftol_history) because everything below the `if` statement is execute once an improvement is observed.


```
if self.swarm.best_cost < best_cost_yet_found:
      relative_measure = self.ftol * (1 + np.abs(best_cost_yet_found))
      LastImprovement = i

      imprvchk = i > MaxStallIter and \
                 np.abs(self.swarm.best_cost - best_cost_yet_found) < relative_measure

      ftol_history.append(imprvchk)

      if all(ftol_history):   
          # stop_reason = "«ftol={}» reached".format(self.ftol)
          # exitflag = 1
          break
      elif self.swarm.best_cost < FitnessLimit:
          # stop_reason = "«fmin={}» reached".format(FitnessLimit)
          # exitflag = 2
          break

  # stall check
  if i - LastImprovement >= MaxStallIter:
      # No improvement in global best for MaxStallIter iterations
      # stop_reason = "«stall_iter={}» reached".format(self.stall_iter)
      break
```

I recommend everyone to look at the current and old implementation and parameters of PSO toolbox.

```
% Default options
options.CognitiveAttraction = 0.5 ;
options.ConstrBoundary = 'penalize' ; 
options.AccelerationFcn = @psoiterate ;
options.DemoMode = 'off' ;
options.Display = 'final' ;
options.FitnessLimit = -inf ;
options.Generations = 200 ;
options.HybridFcn = [] ;
options.InitialPopulation = [] ;
options.InitialVelocities = [] ;
options.KnownMin = [] ;
options.OutputFcns = {} ;
options.PlotFcns = {} ;
options.PlotInterval = 1 ;
options.PopInitRange = [0;1] ;
options.PopulationSize = 40 ;
options.PopulationType = 'doubleVector' ;
options.SocialAttraction = 1.25 ;
options.StallGenLimit = 50 ;
options.StallTimeLimit = Inf ;
options.TimeLimit = Inf ;
options.TolCon = 1e-6 ;
options.TolFun = 1e-6 ;
options.UseParallel = 'never' ;
options.Vectorized = 'off' ;
options.VelocityLimit = [] ;
```

`PopulationSize ` is equivalent to `n_particles`
`Generations ` = `iters`
`TolFun` `ftol`,
`StallGenLimit ` = `stall_iter` or what is called `ftol_iter` at the moment.
`FitnessLimit` = `fmin` that @msat59 has referred to and shown here.

So we can have other features in pyswarms.

pyswarms/base/base_discrete.py
nishnash54: I'm good with the implementation. I just pointed out 2 cases to take into consideration
 1. Changing the default `ftol_iter=50` will break backward compatibility. (This has got nothing to do with any logic)
 2. The `ftol_iter` already handles the stalling part as you have correctly mentioned. But the algorithm is not exactly "stalled" but is not improving enough, so I suggested `n_iter_no_change`.

I'm just going by the understanding from the initial issue where @D4T4-80Y mentioned the use case as
> The stall means no improvement in the best_cost no matter what its value is, which means it **stays constant in two consecutive iterations** . In other word, diff=0 which is always less than ftol.

Again I am fine with the PR as it does not affect the working of `ftol` and `ftol_iter`, but I don't think it solves the issue pointed out in #447 @msat59 Sorry if I wasn't clear about this. :expressionless:

ljvmiranda921: I'll get back to these this weekend, sorry for the delay (got busy with work). Let's be on-track re: Option B that we agreed upon in the original Issue. Also, let's do our best to keep the conversation as calm and objective as possible. cc: @nishnash54 @msat59

msat59: @ljvmiranda921 , we are all good, I think. :wink:

D4T4-80Y:
Any update?

msat59:
still no decision about this?

PR #456, opened 2021-03-31
Installation instructions for Fedora Linux distro users

firefly-cpp:
PySwarms is now available also as a rpm package. It can easily be installed using dnf package manager within Fedora OS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,12 @@ $ cd pyswarms
 $ python setup.py install
 ```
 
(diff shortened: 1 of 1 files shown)


ljvmiranda921 (approved):

ljvmiranda921:
Thanks for this @firefly-cpp !

ljvmiranda921:
@all-contributors please add @firefly-cpp for docs

Issue #484, opened 2021-11-13
Extra documentation for custom objective function


jmmcd:
I found it surprising that the objective function has to accept the entire population (not a single particle) and return one value per particle. My students found it surprising too. In most treatments of optimisation, we define an objective function as a function that accepts a single solution and returns a single real number. 

I see that it's documented at https://pyswarms.readthedocs.io/en/latest/api/pyswarms.utils.functions.html#module-pyswarms.utils.functions.single_obj:

"All objective functions obj_func() must accept a (numpy.ndarray) with shape (n_particles, dimensions). Thus, each row represents a particle, and each column represents its position on a specific dimension of the search-space.

In this context, obj_func() must return an array j of size (n_particles, ) that contains all the computed fitness for each particle."

Ok, I understand that in PySwarms, we are using the vectorised version for speed. I don't disagree with this design decision! Instead, a small addition to the docs would help.

In all the tutorials https://pyswarms.readthedocs.io/en/latest/tutorials.html, the objective is one of the built-in ones, and so the new user is not confronted with the issue. They try to write their own objective function which accepts a single particle, and they don't see any error -- but the algorithm is now not correct! This code runs but only calculates a single objective value for the entire population. I guess that with Numpy broadcasting, the algorithm now operates as if all particles have the same objective value.

```python
import numpy as np
import pyswarms as ps
def my_sphere(x):
    return np.sum(x**2)

options = {'c1': 0.5, 'c2': 0.3, 'w':0.9}
optimizer = ps.single.GlobalBestPSO(n_particles=10, dimensions=2, options=options)
cost, pos = optimizer.optimize(my_sphere, iters=1000)
```

In other cases, depending on the definition of the objective, the user may see a confusing Numpy error message.

## Solutions

1. I would suggest to write a short section "Optimizing your own function" where the objective is hand-written, before the "Optimizing a function with bounds" tutorial section. It could mention this issue and basically use the above example, but show how to use `axis=1` to fix the problem. I could do this and create a PR, what do you think?

2. Could PySwarms make a specific check that the return value from the objective has the expected shape, ie `n_particles`, and write out a user-friendly error message if not? I guess this would be in https://github.com/jmmcd/pyswarms/blob/master/pyswarms/backend/operators.py#L213.

jmmcd:
I implemented these changes here:
https://github.com/jmmcd/pyswarms/

[...]

PR #485, opened 2021-11-15
Extra docs re custom objective function, and a check that objective returns correct shape

jmmcd:
Addresses #484 

## How Has This Been Tested?

Just some manual testing!


ljvmiranda921:
Will do the check! Thanks a lot here @jmmcd !

